PR #45, opened 2015-11-12
Conf files for nsq_* tools

ctdk:
With this PR, the nsq_* tools can at last have a config file specified instead of mangling the upstart configs to supply the needed command line options. Happily the config file isn't required, though, so the existing upstart configs won't stop working unexpectedly between the time the new packages get pushed to packagecloud and the updated chef recipes get deployed.


woodsaj:
@ctdk this looks great. I have also wanted this feature.  Can you add some example configs so it is clear what the format of the config files should be

ctdk:
Sure, I'll turn that around sometime this evening once the kids are in bed and such.

ctdk:
There's a sample ini file, but I just realized that the library for parsing the config options doesn't support specifying the same option multiple times (grrr). I need to fix that before this is done.

ctdk:
This is actually good - the reason it's breaking in circleci is because it's trying to use the old raintank-metric app, and ParseOptions is incompatible between the versions.

PR #183, opened 2016-03-28
use the series T0 attribute

Dieterbe:
no need to store an identical value in chunk.T0


Dieterbe:
made possible by https://github.com/dgryski/go-tsz/pull/15

PR #228, opened 2016-07-01
WIP Kafka

Dieterbe:


metric_tank/in/in.go
```diff
@@ -0,0 +1,76 @@
...
+		msgsAge:           stats.NewMeter(fmt.Sprintf("%s.message_age", input), 0),
+		tmp:               msg.MetricData{Metrics: make([]*schema.MetricData, 1)},
+
+		metrics:  metrics,
+		defCache: defCache,
+		usage:    usage,
+	}
+}
+
+func (in In) Handle(data []byte) {
```
woodsaj: Each message sent to kafka will be a single metric, it wont be a batch of metrics as is done with NSQ.

woodsaj:
@Dieterbe i see a few issues here.

1) messages in kafka will be individual metrics, not batches. 
2) a metricTank instance wont be able to consume from both nsq and kafka, so we should just get rid of all nsq support.   It will be the publishers that provide the transition support by writing to both Kafka and NSQ.

Dieterbe:
> messages in kafka will be individual metrics, not batches. 

OK i will implement this in fake-metrics and MT and try this approach.
Is there a concrete reason why this should be a better approach (other than "it's not needed to batch"). I can see how it should lead to lower delays. Maybe, just maybe also a bit less allocation overhead.  For this style, I'm thinking *not* to add a format uint8 like we do with batched messages, because of the overhead/additional decoding step. We can just have a convention that a given topic will be a given format (eg topic MetricData for msgpacked MetricData messages), and when we optimize the format later, we can just use a different topic.

> a metricTank instance wont be able to consume from both nsq and kafka, so we should just get rid of all nsq support

it **is** able to consume from both. This has proven quite useful when testing in raintank-docker, switching back and forth between nsq and kafka (or both at the same time) using fake-metrics, just to see how the metrics compare (especially allocation rate, GC intervals and "stability" of the metrics-in rate).  I will want to do a similar check in QA. Having one binary that support both makes this testing easier. and in prod it may simplify deployment (or at least keep more options open).
Unless there's a clear disadvantage to keeping the code in, let's remove it later, I want to focus on getting kafka out, tested, and migrated to, first.

Dieterbe:
So i implemented a MetricDataMsgp (MDM) and MetricDataArrayMsgp (MDAM) variant of both fake-metrics producer and MT consumer.
one thing that becomes apparent immediately is that to send a batch of 2000 metrics, which takes a couple ms with NSQ and MDAM, takes +- 360ms with MDM, eg 180 micros per message), this limits the rate of metrics to about 5k/s
I think I read somewhere that we should be able to produce multiple messages at once, maybe in the kafka docs, but i couldn't immediately find how to do this with sarama, where the MDM producer sends each message individually.  this is something we will have to optimize.

Dieterbe:
issues with the producer aside, i also observe, when switching from 5k/s MDM to MDAM:
* mdm has kafka at 100% cpu, mdam has it at 1% (can probably be solved with batching)
* MT cpu usage goes from 60% to 5%
* allocation speed goes from 12MiB/s to 4MiB/s, making GC runs also considerably less frequent with mdam.

i'm having trouble uploading snapshots but here's pics:
pic
pic2


so i'm not sure how to get the MDM approach to work , maybe @woodsaj sees something that I don't. but the good news is that the MDAM approach performs about the same as the nsq approach (in terms of allocations, GC triggers, MT cpu etc) so we can go forward with that.

i've been loading in fake data and the pipeline seems to work fine, so it's time to start looking at rolling out kafka with chef, at least for dev/qa cc @ctdk

Dieterbe:
some more NSQ vs kafka-MDAM observations (this time at 50kmetrics/s):
1) MT cpu stays the same, kafka uses only a few %cpu :+1: 
2) fake-metrics publish time is higher on average but this is not significant.
3) in my devstack i'm often seeing issues only with kafka-mdam that i'm not seeing with nsq:
* fake-metrics publish-to-kafka time shoots >100ms (at which point it drops data that's a design limitation, just to keep in mind when looking at MT ingest). every MT ingest drop that doesn't recover is due to this.
* there are some MT ingest drops that do recover, so they don't correspond to fake-metrics publish time >100ms. it also doesn't seem to correspond to MT GC events. maybe kafka/java GC related.
* MT GC doesn't seem to have an effect on metrics-in (<-> with what we've seen in prod with nsq, maybe because prod has a lot more data)
* long request responses corresponds to ingest drops & fake-metrics overtime
pubtime

huh

so looks like some kind of resource saturation on the host (my laptop)
while these drops/spikes happen cpu does not spike, in fact i noticed a few drops in cpu usage seemingly correlated. there does seem to be a +- 75% correlation with temporary high iowait, but weirdly enough iotop shows that io actually drops during these events.
I don't think this is normal and should be solved, but OTOH if we run some benchmarks on dev/qa we might not have any issues there and have the ability to just move forward.

woodsaj:
> a metricTank instance wont be able to consume from both nsq and kafka

My concern here is that to stop consuming/connecting to nsqd you will have to restart metricTank. This also means we wont be able to turn off nsqd without first restarting all of the metricTank instances.

fake_metrics/out/kafkamdm/kafkamdm.go
```diff
@@ -0,0 +1,87 @@
...
+			return err
+		}
+
+		k.MessageBytes.Value(int64(len(data)))
+
+		prePub := time.Now()
+
+		// We are not setting a message key, which means that all messages will
+		// be distributed randomly over the different partitions.
+		_, _, err = k.client.SendMessage(&sarama.ProducerMessage{
```
woodsaj: use `syncProducer.SendMessages(msgs []*ProducerMessage) error`  to send all of the metrics in a batch instead of individually.

fake_metrics/out/kafkamdm/kafkamdm.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			return err
+		}
+
+		k.MessageBytes.Value(int64(len(data)))
+
+		prePub := time.Now()
+
+		// We are not setting a message key, which means that all messages will
+		// be distributed randomly over the different partitions.
```
woodsaj: We should set a key now (orgId, series Name or metric.Id) so that in future we can shard our metricTank instances.

woodsaj:
In addition to trying syncProducer.SendMessages(),  you will also see significantly better performance using the asynProducer.  https://godoc.org/github.com/Shopify/sarama#AsyncProducer

woodsaj:
Kafka streams messages to clients using the sendfile api, essentially it just gives the client the ability to read the logfile (partition).  By default kafka consumers will only read 32Kb in a single request, or 1 message at a time if the messages are larger then 32kb.  But this can be tuned as shown in https://github.com/raintank/raintank-metric/pull/228/commits/51dec5fc060ab1eabe0fcec62a34b490385789b4

Essentially, the consumer will wait for at least 1Mb of data to be available or for 1second to elapse before reading from the log.  But it will try and read from the log in 4Mb chunks if available.

With this simple change, metricTank was easily able to consume 50K/second metrics with Kafka consuming ~10% CPU.

woodsaj:
While testing on my desktop, the bottleneck was DISK throughput.  After turning on compression in the producer metricTank can now consume 100K metrics/s.

```

Dieterbe:
> My concern here is that to stop consuming/connecting to nsqd you will have to restart metricTank.

Let's do a quick call I think there's a misunderstanding.

Your changes look good. Spent the day strategizing with raj so didn't add anything new. Looking fwd to load test with your changes applied.

Dieterbe:
started to test the changes. looks like a significant improvement. compression takes a bit more cpu and publish duration (not unsuspected), we could make that a tuneable
unfortunately ran into #229 and ran out of time.
also not sure how you've been testing, but https://github.com/raintank/raintank-docker/pull/66 has the commands i use, and some extra goodies.


> In addition to trying syncProducer.SendMessages(), you will also see significantly better performance using the asynProducer. https://godoc.org/github.com/Shopify/sarama#AsyncProducer

the syncProducer is already backed by the asyncProducer but this is probably not how you meant it. 
If you meant using the syncProducer directly then the tsdb service won't be able to tell its clients whether it was able to ingest (i.e. put into kafka) their data, and this seems like a must have feature to me. am i missing something?

also what do you see as advantages of the mdm approach vs mdam (app level chunking) ? I figured to get data more timely (without batching delay) but this no longer applies since 51dec5fc060ab1eabe0fcec62a34b490385789b4

woodsaj:
After playing around with the code lets stick with the syncProducer.  The performance is good enough and as you mentioned it allows us to give clients guarantees that data was successfully received.

I dont think mdam is a viable option as we need to consistantly route metrics to same partition to a) ensure no out of order processing and b) to allow horizontal scaling in future.

Dieterbe:
btw it seems we should be using kafka 0.10, that release comes with a bunch of improvements, and sarama seems to support it (though it's not documented yet). so that's what i'll be testing now

one caveat though:
"The message format in 0.10.0 includes a new timestamp field...producers sending small messages may see a message throughput degradation because of the increased overhead. Likewise, replication now transmits an additional 8 bytes per message."
not a huge deal for us I think, since even when we strip all repetitive metadata, a metricId+value+ts will still be about 20 (or more) +8+4 B

Dieterbe:
* [ ] make sure we can do exactly 50k/s steady
* [ ] make sure no out of order. even when we restart fake-metrics, and when we restart MT

Dieterbe:
i'm going to merge this, obviously it needs more work but that can happen through separate tickets and PR's

PR #291, opened 2016-08-17
Package for debian wheezy

ctdk:
Nothing too crazy, but since graphite-metrictank comes with a wheezy package we may as well have one for metrictank as well (although it's not exactly an ideal configuration).


Dieterbe:
> (although it's not exactly an ideal configuration).

what do you mean with this?

ctdk:
I mean that if you run metrictank on wheezy with sysvinit and it crashes,
it won't come back up is all.

PR #501, opened 2017-01-25
tiny optimization because we know required capacity of slice

replay:


api/query_engine.go
```diff
@@ -113,9 +113,11 @@ func alignRequests(reqs []models.Req, aggSettings []mdata.AggSetting) ([]models.
 	// if we are using raw metrics, we need to find an interval that all request intervals work with.
 	if selected == 0 && len(rawIntervals) > 1 {
 		runTimeConsolidate = true
-		var keys []uint32
+		keys := make([]uint32, len(rawIntervals))
+		i := 0
 		for k := range rawIntervals {
```
Dieterbe: can use `for i, k := range ..` to have i managed for you

DanCech:
Not when ranging over a map though, that would put the key in i and the
value in k wouldn't it?

Dieterbe:
oops. right.

PR #588, opened 2017-04-04
support multiple raw intervals per storage schema.

### LINKED ISSUE #579: match retention polices based on metric interval

woodsaj:
- closes #579
- expand storage schemas into all permutations of the retentions.
eg, a schema with
        retentions=1s:1d,1min:7d,10min:30day
 becomes 3 schemas:
        1 - retentions=1s:1d,1min:7d,10min:30day
        2 - retentions=1min:7d,10min:30day
        3 - retentions=10min:30day
- when calling schemas.Match() pass in the series name and interval.
  We find the schema with a matching pattern and then find
  the sub-schema with the best retention fit. The best fit is when
  the metric interval is >= the rawInterval and less then the interval
  of the next rollup.
  Using our above retention policy, the following matches would occur
  interval=1s:  1s:1d,1min:7d,10min:30day
  interval=10s: 1s:1d,1min:7d,10min:30day
  interval=60s: 1min:7d,10min:30day
  interval=300s: 1min:7d,10min:30day
  interval=3600: 10min:30day

### DIFF
--- a/conf/aggregations.go
+++ b/conf/aggregations.go
@@ -9,16 +9,10 @@ import (
 	"github.com/alyu/configparser"
 )
 
-var DefaultAggregation = Aggregation{
-	Name:              "default",
-	Pattern:           nil,
-	XFilesFactor:      0.5,
-	AggregationMethod: []Method{Avg},
-}
-
 // Aggregations holds the aggregation definitions
 type Aggregations struct {
-	Data []Aggregation
+	Data               []Aggregation
+	DefaultAggregation Aggregation
 }
 
 type Aggregation struct {
@@ -32,6 +26,12 @@ type Aggregation struct {
 func NewAggregations() Aggregations {
 	return Aggregations{
 		Data: make([]Aggregation, 0),
+		DefaultAggregation: Aggregation{
+			Name:              "default",
+			Pattern:           regexp.MustCompile(".*"),
+			XFilesFactor:      0.5,
+			AggregationMethod: []Method{Avg},
+		},
 	}
 }
 
@@ -100,13 +100,13 @@ func (a Aggregations) Match(metric string) (uint16, Aggregation) {
 			return uint16(i), s
 		}
 	}
-	return uint16(len(a.Data)), DefaultAggregation
+	return uint16(len(a.Data)), a.DefaultAggregation
 }
 
 // Get returns the aggregation setting corresponding to the given index
 func (a Aggregations) Get(i uint16) Aggregation {
 	if i+1 > uint16(len(a.Data)) {
-		return DefaultAggregation
+		return a.DefaultAggregation
 	}
 	return a.Data[i]
 }
--- a/conf/schemas.go
+++ b/conf/schemas.go
@@ -8,20 +8,21 @@ import (
 	"strings"
 
 	"github.com/alyu/configparser"
+	"github.com/raintank/metrictank/util"
 )
 
-var DefaultSchema = Schema{
-	Name:       "default",
-	Pattern:    nil,
-	Retentions: Retentions([]Retention{NewRetentionMT(60, 3600*24*7, 120*60, 2, true)}),
+// Schemas contains schema settings
+type Schemas struct {
+	raw           []Schema
+	index         []Schema
+	DefaultSchema Schema
 }
 
-// Schemas contains schema settings
-type Schemas []Schema
+type SchemaSlice []Schema
 
-func (s Schemas) Len() int           { return len(s) }
-func (s Schemas) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
-func (s Schemas) Less(i, j int) bool { return s[i].Priority >= s[j].Priority }
+func (s SchemaSlice) Len() int           { return len(s) }
+func (s SchemaSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+func (s SchemaSlice) Less(i, j int) bool { return s[i].Priority >= s[j].Priority }
 
 // Schema represents one schema setting
 type Schema struct {
@@ -31,8 +32,40 @@ type Schema struct {
 	Priority   int64
 }
 
-func NewSchemas() []Schema {
-	return nil
+func NewSchemas(schemas []Schema) Schemas {
+	s := Schemas{
+		raw: schemas,
+		DefaultSchema: Schema{
+			Name:       "default",
+			Pattern:    regexp.MustCompile(".*"),
+			Retentions: Retentions([]Retention{NewRetentionMT(1, 3600*24*1, 600, 2, true)}), // 1s:1day:10mim:2:true
+		},
+	}
+	s.BuildIndex()
+	return s
+}
+
+func (s *Schemas) BuildIndex() {
+	s.index = make([]Schema, 0)
+	for _, schema := range s.raw {
+		for pos := range schema.Retentions {
+			s.index = append(s.index, Schema{
+				Name:       schema.Name,
+				Pattern:    schema.Pattern,
+				Retentions: schema.Retentions[pos:],
+				Priority:   schema.Priority,
+			})
+		}
+	}
+	// add the default schema
+	for pos := range s.DefaultSchema.Retentions {
+		s.index = append(s.index, Schema{
+			Name:       s.DefaultSchema.Name,
+			Pattern:    s.DefaultSchema.Pattern,
+			Retentions: s.DefaultSchema.Retentions[pos:],
+			Priority:   s.DefaultSchema.Priority,
+		})
+	}
 }
 
 // ReadSchemas reads and parses a storage-schemas.conf file and returns a sorted schemas structure
@@ -40,15 +73,15 @@ func NewSchemas() []Schema {
 func ReadSchemas(file string) (Schemas, error) {
 	config, err := configparser.Read(file)
 	if err != nil {
-		return nil, err
+		return Schemas{}, err
 	}
 
 	sections, err := config.AllSections()
 	if err != nil {
-		return nil, err
+		return Schemas{}, err
 	}
 
-	var schemas Schemas
+	var schemas []Schema
 
 	for i, sec := range sections {
 		schema := Schema{}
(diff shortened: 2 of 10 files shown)


conf/aggregations.go
```diff
@@ -9,16 +9,10 @@ import (
...
-	Pattern:           nil,
-	XFilesFactor:      0.5,
-	AggregationMethod: []Method{Avg},
-}
-
 // Aggregations holds the aggregation definitions
 type Aggregations struct {
-	Data []Aggregation
+	Data               []Aggregation
+	DefaultAggregation Aggregation
```
Dieterbe: the point of the default is that it applies irrespective of what your Aggregation definitions are, so why move it from global to a member?

woodsaj: DefaultAggregation is only modified by unitTests.  Keeping it as part of the Aggregations{} instance makes it easier for unit tests to modify the defaults for testing purposes.

eg, calling mdata.SetSingleAgg(met ...conf.Method) wont break other unit tests that may be exacting the DefaultAggregation to be set to the defaults.

conf/schemas.go
```diff
@@ -58,50 +91,114 @@ func ReadSchemas(file string) (Schemas, error) {
...
+
+	return NewSchemas(schemas), nil
 }
 
 // Match returns the correct schema setting for the given metric
 // it can always find a valid setting, because there's a default catch all
 // also returns the index of the setting, to efficiently reference it
-func (s Schemas) Match(metric string) (uint16, Schema) {
-	for i, schema := range s {
+func (s Schemas) Match(metric string, interval int) (uint16, Schema) {
```
Dieterbe: I find this new function hard to understand.
One reason I think is because it's a mashup between two mental models:
A) all our schemas/retentions as 1 long list
B) a list of entries (based on pattern), and then each entry has a list of retention lists (due to the new "permutation" stuff). i.e. a nested structure.

This function has a few places where the logic works in the domain of B (e.g. taking the first retention of a given pattern, then working with `len(schema.Retentions)`, or iterating that list which actually represents items further down in the index list. I drafted some refactoring, but couldn't come up with anything better. I think a nested, B-style model would be nicer for our index datastructure, but then we would need two identifiers to represent a schema, which is to be avoided.

conf/schemas.go
```diff
@@ -58,50 +91,114 @@ func ReadSchemas(file string) (Schemas, error) {
...
+				return uint16(i), *schema
+			}
+			// search through the retentions to find the first one where
+			// the metric interval is < SecondsPerPoint of the retention.
+			// The schema is then the previous retention.
+			for j, ret := range schema.Retentions {
+				if interval < ret.SecondsPerPoint {
+					// if there are no retentions with SecondsPerPoint >= interval
+					// then we need to use the first retention. Otherwise, the retention
+					// we want to use is the previous one.
```
Dieterbe: this is confusing. we branch if interval < SecondsPerPoint but this talks about SecondsPerPoint >= interval which means interval <= SecondsPerPoint

conf/schemas.go
```diff
@@ -58,50 +91,114 @@ func ReadSchemas(file string) (Schemas, error) {
...
+					if j > 0 {
+						j--
+					}
+					pos := i + j
+					return uint16(pos), *(s.index[pos])
+				}
+			}
+			// no retentions found with SecondsPerPoint > interval. So lets just use the retention
+			// with the largest secondsPerPoint.
+			pos := len(schema.Retentions) - 1 + i
```
Dieterbe: is it just me or would `i + len(schema.Retentions) - 1` be easier to understand? i think of it like "i is the index where we're at, so that's the base, then from there jump to the last retention" so this order makes more sense to me.

conf/schemas.go
replay: I'd like the suggestion B more mainly for performance reasons. If i understand it right in the current solution let's say if we have `3` patterns and each has `3` retentions, then if something calls `Match()` with a metric/interval that matches the last retention of the last pattern, then `.MatchString()` will be called `9` times because it's called for each retention of each pattern. Assuming that pattern matching is relatively expensive compared to an index lookup it would seem more efficient to have a nested structure that allows us to only call `.MatchString()` once per pattern.

replay: I agree that it's kind of hard to wrap ones head around that concept, but it's pretty cool that in the end that pattern/retention can be looked up with only one index.
Maybe it would be good to have some comment that illustrates how that works, I imagine the `s.index` slice to look like a partitioned list of retentions where each partition is a pattern:
```
|-----------------------------------------------------------------------------|
|   pattern1          |  pattern2   |            pattern3                     |
|-----------------------------------------------------------------------------|
| ret1 | ret2 | ret3 | ret4 | ret5 | ret6 | ret7 | ret8 | ret9 | ret10 |
|-----------------------------------------------------------------------------|

1) if metric matches patternX -> find the best retention within patternX
2) otherwise -> skip len(patternX.retentions) and back to 1)
```

conf/schemas.go
```diff
@@ -58,50 +91,114 @@ func ReadSchemas(file string) (Schemas, error) {
...
-		return DefaultSchema
+	if i+1 > uint16(len(s.index)) {
+		return s.DefaultSchema
+	}
+	return *(s.index[i])
+}
+
+// TTLs returns a slice of all TTL's seen amongst all archives of all schemas
+func (schemas Schemas) TTLs() []uint32 {
+	ttls := make(map[uint32]struct{})
```
replay: Couldn't this nested loop and the following loop all be replaced if instead `schemas.index` is used?

conf/schemas.go
```diff
@@ -58,50 +91,114 @@ func ReadSchemas(file string) (Schemas, error) {
...
+	var ttlSlice []uint32
+	for ttl := range ttls {
+		ttlSlice = append(ttlSlice, ttl)
+	}
+	return ttlSlice
+}
+
+// MaxChunkSpan returns the largest chunkspan seen amongst all archives of all schemas
+func (schemas Schemas) MaxChunkSpan() uint32 {
+	max := uint32(0)
```
replay: Same here... Only one loop would be necessary by indexing over `schemas.index`.

conf/schemas.go
Dieterbe: it could. this code predates the "permutation" stuff. There's no need for caching since this function rarely gets called and is not in any performance critical path.

replay: K, true. only called once in fact

conf/aggregations.go
Dieterbe: but both SetSingleAgg and SetSingleSchema modify global state. wouldn't that global state stay just as "dirty" as it was before? Your goal is noble but I think it can only be fulfilled by passing the Schemas and Aggregations around instead of having them as globals in the mdata package.

conf/schemas.go
Dieterbe: to be clear I'm not saying we should change the structure, I also really like the single index.  I just think it's important to make this logic as simple as we can.

conf/schemas_test.go
```diff
@@ -0,0 +1,137 @@
...
+		So(len(ttls), ShouldEqual, 5)
+	})
+}
+func TestMaxChunkSpan(t *testing.T) {
+	schemas := schemasForTest()
+	Convey("When getting maxChunkSpan", t, func() {
+		max := schemas.MaxChunkSpan()
+		So(max, ShouldEqual, 60*60*6)
+	})
+}
```
Dieterbe: thanks for adding unit tests for my code :D :+1:

conf/schemas_test.go
```diff
@@ -0,0 +1,137 @@
...
+			So(schema.Name, ShouldEqual, "a")
+			So(schema.Retentions[0].SecondsPerPoint, ShouldEqual, 10)
+		})
+		Convey("When metric has 30s raw interval", func() {
+			id, schema := schemas.Match("a.foo", 30)
+			So(id, ShouldEqual, 0)
+			So(schema.Name, ShouldEqual, "a")
+			So(schema.Retentions[0].SecondsPerPoint, ShouldEqual, 10)
+		})
+		Convey("When metric has 1s raw interval", func() {
```
Dieterbe: any particular reason behind the ordering 10s -> 30s -> 1s ? i would do them in ascending order.

conf/schemas.go
```diff
@@ -8,20 +8,21 @@ import (
...
 )
 
-var DefaultSchema = Schema{
-	Name:       "default",
-	Pattern:    nil,
-	Retentions: Retentions([]Retention{NewRetentionMT(60, 3600*24*7, 120*60, 2, true)}),
+// Schemas contains schema settings
+type Schemas struct {
+	raw           []Schema
+	index         []*Schema
```
woodsaj: at one stage match was performing a range over the index.  A range of a slice copies the values, so only copying the reference is much faster then copying the whole schema struct.

But doenst look like this is needed anymore

conf/schemas.go
woodsaj: yes, this function is complex. It took a long time to come to this solution that only requires a single call to matchString(), but still allows a single schema ID to be passed around as a reference.

I think @replay's illustration will go a long way to clarifying what is going on.  But i would change it to

image

conf/schemas.go
woodsaj: It is just you.  This order makes more sense to me.

conf/aggregations.go
woodsaj: SetSingleAgg and SetSingleSchema only modify the state within the "mdata" pacakge. Allowing them to modify global variables in other packages is a recipe for disaster.

conf/schemas.go
woodsaj: This is just basic logic.  if a < b is true, then b >= a is false.

ie these are the same thing
`if interval < ret.SecondsPerPoint`
`if !(ret.SecondsPerPoint >= interval)`

Dieterbe:
I think we shouldn't use the word permutation here. a permutation is [something else](https://en.wikipedia.org/wiki/Permutation) (e.g. implies reordering) and it's confusing.
I can't find word a perfectly fitting word but "subsequence" is at least closer.

Dieterbe:
Isn't the whole point of the createmissing to add the aggmetrics structure when this instance has no knowledge of them? So that a new instance can start up, process the metricpersist backlog, and know how far metrics (that it hasn't seen yet, but will, soon) have been saved yet.  see e03d2fdc696001ebdab2d82a279075c416d9742a and #485 ?
by ignoring messages if they're not in the index, doesn't that defeat the whole point?

woodsaj:
No.  The index is always loaded into memory first before anything else.  The only time a savedChunk message can be received and there be no entry in the index is if the metric has been deleted from the index.  In which case we dont need to worry about the saveChunk anymore.

woodsaj:
Actually there is a second reason.  The savedChunk message is for a metric that is on a different partition to the partitions being handled by this instance. In which case we also want to ignore the message.

* NotifierKafka is partition aware, so only instances only receive savedChunk messages for metrics in the partitions they are processing.  But notifierNSQ just broadcasts savedChunk messages to all nodes.

Dieterbe:
ok makes sense. for the record I just added CreateMissing to notifierNSQ because it seemed like the right thing to do. I didn't put a whole lot of thought into it.

Also, a nice property of the current index code is that while metricdef updates (e.g. existing defs) are subject to update-interval, new metrics will have archive.LastSave=0, and trigger an immediate save to cassandra.
This means that even in cases where the metric is relatively new (e.g. much newer than index update-interval) (but new enough to trigger a chunk save), and a new node comes online and gets a savedChunk for that metric, we'll have done our best to update the cassandra index asap, and the chance of a legitimate miss in the index is so low that an extra chunk save wouldn't matter.

conf/schemas.go
Dieterbe: > if a < b is true, then b >= a is false.

no, these are almost equivalent. e.g. if a is 1 and b is 2 then both are true.

> ie these are the same thing

no they are not:
```
	interval := 10
	SecondsPerPoint := 20
	fmt.Println(interval < SecondsPerPoint)
	fmt.Println(!(SecondsPerPoint >= interval))
```

this prints:
```
true
false
```

PR #625, opened 2017-05-03
Expr2 awoods

woodsaj:
- simplify the graphiteFunc interface to a single "Plan" method that
  validates arguments, modifies request options if needed (to,from, consolidateBy)
   and returns a closure (pointer to an execution function).
  By using a closure, we can pull the serieslist from dependent functions
  before starting our execution phase.  This also works well with functions that accept
  multiple serieslists, as they can be merged into a single serieslist before executing.

- use separate plans for each target in the request. The request to,from
  are scoped to a specific plan. If there are multiple targets and one has a function
  that needs to modify the to,from timestamps, it should not affect other targets.


expr/func_alias.go
```diff
@@ -1,38 +1,50 @@
...
-func (s FuncAlias) Signature() ([]argType, []argType) {
-	return []argType{seriesList, str}, []argType{seriesList}
-}
+// alias(seriesList, newName)
+func (s FuncAlias) Plan(args []*expr, namedArgs map[string]*expr, plan *Plan) (execHandler, error) {
+	// Validate arguments ///
+	if len(args) > 2 || len(namedArgs) > 0 {
+		return nil, ErrTooManyArg
+	}
+	if len(args) < 1 {
```
DanCech: shouldn't this be `< 2`?

PR #656, opened 2017-06-16
Allow copy/pasta for cassandra config

shanson7:
Fix duplicate word so command doesn't fail on copy paste. Also change datacenter to the cassandra default, for test setup ease.


woodsaj:
thanks @shanson7

PR #709, opened 2017-08-08
Tracing support

Dieterbe:
a
b


woodsaj:
This is looking good @Dieterbe 
one question, are we going to be able to correlate the remote calls back to the originating caller.  ie
for a render call that times out in GetTargetsRemote, are we going to be able to dive in and see a trace of getData on each MT instance?

api/middleware/middleware.go
```diff
@@ -37,6 +37,9 @@ func getOrg(req *http.Request, multiTenant bool) (int, error) {
 	}
 	orgStr := req.Header.Get("x-org-id")
 	if orgStr == "" {
+		orgStr = req.Header.Get("x-grafana-org-id")
+	}
+	if orgStr == "" {
```
Dieterbe: @woodsaj thoughts on this? I did it because grafana uses that header now, and to make the dev stack work where grafana talks to MT directly

woodsaj: I am ok with this.  We can also pass this through graphite-web if you like, just need to update the "REMOTE_STORE_FORWARD_HEADERS in local_settings.py

https://github.com/raintank/deployment_tools/blob/master/metrictank/image/graphite-web/local_settings.py#L135

DanCech: x-grafana-org-id is just going to be the internal id of the org within the grafana instance, it won't have any correlation with the grafana.com org id

Dieterbe:
> one question, are we going to be able to correlate the remote calls back to the originating caller. ie
for a render call that times out in GetTargetsRemote, are we going to be able to dive in and see a trace of getData on each MT instance?

even better I think. the way this is supposed to work is that the trace context gets sent along with the getData RPC call, so the peer MT can effectively add its own logs and trace items into the "parent" context from the caller, and it'll show up as 1 tree in the jaeger UI.

i still have to get this to work, but that's how it should be.

Dieterbe:
here's a screenie of the current implementation, demonstrating the above.
bonus : i added in a fake error to make it more interesting
remote-err

api/models/node.go
```diff
@@ -37,11 +42,28 @@ type IndexFind struct {
...
 type GetData struct {
 	Requests []Req `json:"requests" binding:"Required"`
 }
 
+func (g GetData) Trace(span opentracing.Span) {
+	for _, r := range g.Requests {
+		r.TraceLog(span)
+	}
+}
+
```
Dieterbe: i suspect this may be overdoing it. for every request in a GetData RPC call, log all the request properties. this could be too much overhead

api/response/response.go
```diff
@@ -27,6 +31,14 @@ func Write(w http.ResponseWriter, resp Response) {
 	return
 }
 
+func WriteErr(ctx context.Context, w http.ResponseWriter, resp Response) {
```
woodsaj: every call to WriteErr is passing `writeErr(ctx.Req.Context(), ctx, error)`
so why not just use 
```
func WriteErr(c *macaron.Context, resp Response) {
  ctx := c.Req.Context()
  .....
}
```
It may even be a better idea to just include the error logger in the Write() method, eg.
```
func Write(ctx *macaron.Context, resp Response) {
   defer resp.Close()
	body, err := resp.Body()
	if err != nil {
		w.WriteHeader(http.StatusInternalServerError)
		w.Write([]byte(err.Error()))
		return
	}
	for k, v := range resp.Headers() {
		w.Header().Set(k, v)
	}
	w.WriteHeader(resp.Code())
	w.Write(body)
        
        // log errors
        if errResp, ok := resp.(Error); ok { // or maybe "if resp.Code() >= 400"
            span := opentracing.SpanFromContext(ctx)
 	    body, _ := resp.Body()
 	    span.LogFields(log.String("error.kind", string(body)))
 	     tags.Error.Set(span, true)
    }
}
```

conf/tracing.go
```diff
@@ -0,0 +1,38 @@
+package conf
```
woodsaj: why is this package called "conf" when all handles is tracing?

Dieterbe: conf is not new. https://github.com/raintank/metrictank/tree/master/conf
before it was just used for extra config related things like storage-schemas and aggregation rules config, and now also for tracing config/setup.  I thought about putting it in a `tracing` package but it would be a tiny package. though perhaps that's the better way.

api/response/response.go
Dieterbe: I liked that the response package was pretty standalone and not tied into macaron, and wanted to keep it that way.

PR #742, opened 2017-10-04
Add maxSeries function

shanson7:
This code is mostly copy/pasta from sumSeries.

I jumped directly to stable since the code is rather simple. I spot-checked this against graphite and the only difference I noticed was some double-precision differences (e.g. graphite says `99.8005834878` and MT says `99.806`, which is was MT returned in the raw render as well).


shanson7:
Any love?

PR #792, opened 2017-12-19
fix seriesByTag pathExpression

DanCech:


replay:
looks good, i'm deploying it to the test env

Issue #887, opened 2018-04-12
debian package are not installable (tried on jessie and stretch)


beorn-:
It triggers the "version number does not start with digit" error

the version is a hash which is forbidden by debian policy, the version in the debian/control file has to be in the form : 0.8.3-1

where 0.8.3 is the version and -1 is the build revision

my opinion is that you could use latest tagged version and increment an int for the revision it's a common practice

woodsaj:
I dont think this is a problem with our build scripts. Though we only use the docker images these days, in the past we have used the deb packages.

My guess is that there was problem during a build and the `git describe` only returned the commit id and not the tag information.

woodsaj:
confirmed, i found broken packages in the repo from 6months ago.
eg `metrictank_f2df73a_amd64.deb`

woodsaj:
ok, i have deleted the bad packages from the repo. Should be all good now.

PR #889, opened 2018-04-13
Version was invalid for dpkg based packages used tag-commitssincetag

beorn-:
This should fix issue #887 

the generated string for version was : 0.8.3-113-gef6c1192

According to debian documentation 

_The version number format is: [epoch:]upstream_version[-debian_revision]_

which would give here 0.8.3-113


woodsaj:
This is not needed.  the issue described in #887 was simply due to bad packages being added to the repo 6months ago.  These have been removed.

PR #919, opened 2018-05-20
Timestamp fixes etc 

Dieterbe:
trying to create new PR because circleCI keeps saying a test failed but i don't see how/why


Dieterbe:
this replaces #915 and #916

PR #945, opened 2018-06-14
Use atomic operations and a read lock instead of a write lock

shanson7:
This PR was an attempt to reduce the exclusive lock section in the ingest path. 

The idea is that the map isn't truly being modified, so we don't need to hold a write lock. The behavioral change is that if the same point is `Update`d by two threads, the partition/LastUpdate is not guaranteed to match. In practice, I believe that `LastUpdate` should pretty much be near realtime (and is mostly heuristic anyway). The `partition` shouldn't change frequently anyway and should be eventually consistent.

Similar changes could be made to `AddOrUpdate` (optimistically acquiring a read lock) but I wasn't sure how many calls to `AddOrUpdate` actually resulted in a write.

In our setup, we saw a 30%-40% bump in our backlog processing from this change.

### DIFF
--- a/idx/memory/memory.go
+++ b/idx/memory/memory.go
@@ -219,12 +219,11 @@ func (m *MemoryIdx) Update(point schema.MetricPoint, partition int32) (idx.Archi
 
 	existing, ok := m.defById[point.MKey]
 	if ok {
-		oldPart := existing.Partition
 		if LogLevel < 2 {
 			log.Debug("metricDef with id %v already in index", point.MKey)
 		}
 
-		if atomic.LoadInt64(existing.LastUpdate) < int64(point.Time) {
+		if atomic.LoadInt64(&existing.LastUpdate) < int64(point.Time) {
 			atomic.SwapInt64(&existing.LastUpdate, int64(point.Time))
 		}
 		oldPart := atomic.SwapInt32(&existing.Partition, partition)
@@ -247,12 +246,11 @@ func (m *MemoryIdx) AddOrUpdate(mkey schema.MKey, data *schema.MetricData, parti
 
 	existing, ok := m.defById[mkey]
 	if ok {
-		oldPart := existing.Partition
 		log.Debug("metricDef with id %s already in index.", mkey)
 		if atomic.LoadInt64(&existing.LastUpdate) < int64(data.Time) {
 			atomic.SwapInt64(&existing.LastUpdate, int64(data.Time))
 		}
-		existing.Partition = partition
+		oldPart := atomic.SwapInt32(&existing.Partition, partition)
 		statUpdate.Inc()
 		statUpdateDuration.Value(time.Since(pre))
 		m.RUnlock()


Dieterbe:
i haven't looked in depth at this yet, but FWIW I'm pretty sure you can't just do atomic writes to integers while elsewhere reading (even when holding the read lock).
i recently read an interesting article that explained some of this stuff (but not sure if was exactly that), it talked about instruction reordering and how there's much less memory guarantees than some people assume. I couldn't find the article but I did find https://github.com/golang/go/issues/5045#issuecomment-252730563 : 

> I'm fairly certain the rules will be that Go's atomics guarantee sequential consistency among the atomic variables (behave like C/C++'s seqconst atomics), and that you shouldn't mix atomic and non-atomic accesses for a given memory word.

we may want to pursue tracking of partition/lastUpdate in a separate structure that is entirely based on atomics, just thinking out loud.

shanson7:
It seems that it would be easy enough to create an AtomicUint32 (and the whole family) to encapsulate these actions. The trick is that the MetricDefinition would need to use it all over...

Dieterbe:
> The trick is that the MetricDefinition would need to use it all over...

related: the in-memory index (including its types and MetricDefinition) are due for a make-over anyway.
they're quite inefficient.

shanson7:
So, we were running the "unsafe" version of this (mixing atomic/non-atomic accesses) for over a month and didn't see issues, but I recently rolled out a new build without this branch as I didn't want race conditions in there, and I wasn't ready to commit to this change. However, without this change I saw about a 30% slowdown in backfill processing without it. So, I decided to just bite the bullet and look through all the `.Partition` and `.LastUpdate` accesses and access them atomically (unless a write lock is already held).

shanson7:
Opened #969 for failed test

Dieterbe:
did deploying this PR but with these atomic reads added give you your 30% speedup back?

idx/memory/memory.go
```diff
@@ -213,20 +214,19 @@ func (m *MemoryIdx) Stop() {
...
 	if ok {
-		oldPart := existing.Partition
 		if LogLevel < 2 {
 			log.Debug("metricDef with id %v already in index", point.MKey)
 		}
 
-		if existing.LastUpdate < int64(point.Time) {
-			existing.LastUpdate = int64(point.Time)
+		if atomic.LoadInt64(&existing.LastUpdate) < int64(point.Time) {
+			atomic.SwapInt64(&existing.LastUpdate, int64(point.Time))
```
Dieterbe: note that this is racey.
let's say existing.LastUpdate is very old (30 days ago)
then a point comes in for 29 days ago, and concurrently another one for a day ago via a different kafka partition, and then no more points.
in that case, we can have concurrent Update calls, resulting in the LastUpdate field being updated to 29 days ago, but never to a day ago.
note that for any kafka given partition, carbon stream or prometheus POST we never have overlap in our Update calls.

so in practice, doesn't seem like an issue, but perhaps we should document this something under "tradeoffs and extremely rare edge cases" or something.

Dieterbe: or we can solve it by either:
1) doing CompareAndSwap in a loop until we're able to swap for the value we wanted to swap
2) confirming the swapped out value (return value of SwapInt64) is smaller than what we swapped in. if not, put the old value back, check that we didn't swap for an ever higher value (placed by a concurrent Update call), etc

shanson7:
Yes, this PR definitely got us back where we wanted to be. Ingest rate is not very consistent, but we average ~90k dp/s/core (on 8 cores, we see spikes up to 1M dp/s but average ~700k). Without this change, we were barely breaking 400k dp/s. We have been running this in production for about 2 months now, with no noticeable issues.

I really look forward to seeing if it benefits your speeds as well (the trade-off, I suppose, is greater CPU usage during ingest).

idx/memory/memory.go
Dieterbe: I don't understand this. if you use a write lock, you can lock, check value, if we have a larger one, update, unlock. this works regardless of the order between two concurrent update operations. (the most recent will always survive).

I think my proposal above will also solve it, and at almost no additional cost.

shanson7: I was extending it to the `Partition` update as well. Perhaps `Partition` should only be updated when we have a newer timestamp as well?

Dieterbe: the partition property is really only to shard the index in cassandra, so nodes on startup know which metrics they're responsible for.
i'm not sure if we even properly support live partition changes of metrics (i.e. whether after the change we properly display both the old and new data)
under concurrent updates it's probably ok for the partition to be ambiguous ("under transition"), but once update operations are serialized, the later one should probably always win, even when data is sent out of order. I think MT's behavior in these scenarios is so undefined that probably either way works.

Dieterbe:
I have
1) rebased your work on top of master
2) added a commit with what I think is the solution to update LastUpdate properly. I think it should not be any slower (or more complicated) then what you had. 
3) i imagine my function would be inlined but for some reason couldn't prove it via `go build -gcflags '-m'` so i also added a commit manually inlining it (but i don't intend to merge that commit)

here's my branch: https://github.com/grafana/metrictank/tree/readLockUpdates

then i filled up kafka with some MetricData data and tested ingestion with each version (twice)

```
Mon Aug 27 11:08:00 CEST 2018 running mt-dieter
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
Mon Aug 27 11:10:00 CEST 2018 running mt-dieter
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
/home/dieter/go/src/github.com/grafana/metrictank
Mon Aug 27 11:12:00 CEST 2018 running mt-dieter-inline
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
Mon Aug 27 11:14:01 CEST 2018 running mt-dieter-inline
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
/home/dieter/go/src/github.com/grafana/metrictank
Mon Aug 27 11:16:02 CEST 2018 running mt-master
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
Mon Aug 27 11:18:02 CEST 2018 running mt-master
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
/home/dieter/go/src/github.com/grafana/metrictank
Mon Aug 27 11:20:03 CEST 2018 running mt-sean
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
Mon Aug 27 11:22:03 CEST 2018 running mt-sean
Restarting docker-dev-custom-cfg-kafka_metrictank_1 ... done
/home/dieter/go/src/github.com/grafana/metrictank
```
https://snapshot.raintank.io/dashboard/snapshot/hMzSp4LGcBaJ5iKDrvrWMueMWzkxTUL6?orgId=2
https://snapshot.raintank.io/dashboard/snapshot/OCgXyoe0RQEURuAS6gV3Iuzo3K7c5qJS?orgId=2

cpu difference looks fine (tiny)
version sean, dieter and dieter-inline all seem to perform the same, so i suggest we merge my branch minus the inline commit.
important to note that this is not under a http workload, so any gains here are the baseline gains, under concurrency we can expect to gain more

shanson7:
Yeah, looks great to me! I'm excited to see if you see a difference in ingest speed as pronounced as we did.

Dieterbe:
rollout to our internal monitoring environment..
backfill speeds of different instances as the pods restart one by one.
left: old, right: new.
note increased speeds and also shorter cluster restart time (both charts are exactly 7h wide)
a

PR #959, opened 2018-07-09
Add isNonNull function

stivenbb:
Implementation of isNonNull() graphite function.

Takes a seriesList and counts up non-null values. Returns a seriesList where 1 is specified for non-null values, and 0 is specified for null values.

### DIFF
--- a/expr/func_isnonnull.go
+++ b/expr/func_isnonnull.go
@@ -36,11 +36,14 @@ func (s *FuncIsNonNull) Exec(cache map[Req][]models.Series) ([]models.Series, er
 		transformed := models.Series{
 			Target:       fmt.Sprintf("isNonNull(%s)", serie.Target),
 			QueryPatt:    fmt.Sprintf("isNonNull(%s)", serie.QueryPatt),
-			Tags:         serie.Tags,
+			Tags:         make(map[string]string),
 			Datapoints:   pointSlicePool.Get().([]schema.Point),
 			Interval:     serie.Interval,
 			Consolidator: serie.Consolidator,
 			QueryCons:    serie.QueryCons,
+		}
+		for k, v := range serie.Tags {
+			transformed.Tags[k] = v
 		}
 		transformed.Tags["isNonNull"] = "1"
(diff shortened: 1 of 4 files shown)


expr/func_isnonnull.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (s *FuncIsNonNull) Exec(cache map[Req][]models.Series) ([]models.Series, error) {
+	series, err := s.in.Exec(cache)
+	if err != nil {
+		return nil, err
+	}
+
+	out := make([]models.Series, 0, len(series))
+	for _, serie := range series {
+		transformed := models.Series{
```
replay: the `out` slice is already allocated with the correct size. wouldn't it be better to just update its properties instead of instantiating a new` models.Series` and then copying it into `out`?
f.e.:
`out[i].Target = fmt.Sprintf("isNonNull(%s)", serie.Target),` (`i` is the index of the loop)

expr/func_isnonnull.go
```diff
@@ -0,0 +1,62 @@
...
+			transformed.Tags[k] = v
+		}
+		transformed.Tags["isNonNull"] = "1"
+		for _, p := range serie.Datapoints {
+			if math.IsNaN(p.Val) {
+				p.Val = 0
+			} else {
+				p.Val = 1
+			}
+			transformed.Datapoints = append(transformed.Datapoints, p)
```
replay: `transformed.Datapoints` could also be preallocated and then just updated.

stivenbb:
Added requested change

replay:
GitHub seems to be broken right now. I clicked the "Merge" button which ended up in an error page. Looking at the `master` branch it seems the PR got merged but this PR page did not get updated accordingly.

replay:
@stivenbb  I'm afraid I'll only be able to just close this PR. As the commits are merged already, I guess that shouldn't make a difference:
screenshot from 2018-07-19 16-27-54

shanson7:
I see the commits in there, and @stivenbb gets credit as a contributor :)

I think closing this is fine.

PR #966, opened 2018-07-26
Add asPercent function

stivenbb:
Native implementation of asPercent() Graphite function. (http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.asPercent)

Added a new argument type `ArgIn` that allows multiple other argument types. This was necessary for the *total* argument. Some of the code borrowed from an abandoned PR: https://github.com/grafana/metrictank/pull/672

In terms of speed improvement:
```
---------- Native Implementation ----------
Requests      [total, rate]            900, 5.01
Duration      [total, attack, wait]    3m0.13756s, 2m59.799999s, 337.561ms
Latencies     [mean, 50, 95, 99, max]  72.006704ms, 38.065ms, 342.887ms, 472.467ms, 765.657ms
Bytes In      [total, mean]            130300948, 144778.83
Status Codes  [code:count]             200:900
Error Set:
---------- Graphite (Python) Implementation ----------
Requests      [total, rate]            900, 5.01
Duration      [total, attack, wait]    3m6.337648s, 2m59.799999s, 6.537649s
Latencies     [mean, 50, 95, 99, max]  797.282367ms, 167.489ms, 4.789024s, 6.756318s, 8.006429s
Bytes In      [total, mean]            144407224, 160452.47
Status Codes  [code:count]             200:900
Error Set:
```

On average, the native implementation was 11x faster, median was 4x faster, p95 was 14x faster, p99 was 14x faster and max was over 10x faster


expr/func_aspercent.go
```diff
@@ -0,0 +1,241 @@
...
+		if len(series) == 1 {
+			totalsSerie.Target = fmt.Sprintf("sumSeries(%s)", totalsSerie.Target)
+			totalsSerie.QueryPatt = fmt.Sprintf("sumSeries(%s)", totalsSerie.QueryPatt)
+			totalsSerie.Tags = map[string]string{"name": totalsSerie.Target}
+		}
+	} else if totals != nil {
+		if len(totals) == 1 {
+			totalsSerie = totals[0]
+		} else if len(totals) == len(series) {
+			sort.Slice(series, func(i, j int) bool {
```
shanson7: Maybe a comment about why there are sorted.

expr/func_aspercent.go
```diff
@@ -0,0 +1,241 @@
...
+	return context
+}
+
+func (s *FuncAsPercent) Exec(cache map[Req][]models.Series) ([]models.Series, error) {
+	series, err := s.in.Exec(cache)
+	if err != nil {
+		return nil, err
+	}
+
+	var outSeries []models.Series
```
shanson7: Could you do a 

```
defer cache[Req{}] = append(cache[Req{}], outSeries...)
```

here instead of all the individual cache steps?

expr/func_aspercent.go
```diff
@@ -0,0 +1,241 @@
...
+					for i := range serie1.Datapoints {
+						serie1.Datapoints[i].Val = math.NaN()
+					}
+				} else {
+					serie2 := totalSeries[key]
+					serie1.QueryPatt = fmt.Sprintf("asPercent(%s,%s)", serie1.QueryPatt, serie2.QueryPatt)
+					serie1.Target = fmt.Sprintf("asPercent(%s,%s)", serie1.Target, serie2.Target)
+					serie1.Tags = map[string]string{"name": serie1.Target}
+					for i := range serie1.Datapoints {
+						serie1.Datapoints[i].Val = computeAsPercent(serie1.Datapoints[i].Val, serie2.Datapoints[i].Val)
```
shanson7: I think this violates https://github.com/grafana/metrictank/blob/master/expr/NOTES#L35

expr/func_aspercent.go
```diff
@@ -0,0 +1,241 @@
...
+			sort.Slice(totals, func(i, j int) bool {
+				return totals[i].Target < totals[j].Target
+			})
+			for i, serie1 := range series {
+				serie2 := totals[i]
+				serie1.QueryPatt = fmt.Sprintf("asPercent(%s,%s)", serie1.QueryPatt, serie2.QueryPatt)
+				serie1.Target = fmt.Sprintf("asPercent(%s,%s)", serie1.Target, serie2.Target)
+				serie1.Tags = map[string]string{"name": serie1.Target}
+				for i := range serie1.Datapoints {
+					serie1.Datapoints[i].Val = computeAsPercent(serie1.Datapoints[i].Val, serie2.Datapoints[i].Val)
```
shanson7: I think this violates https://github.com/grafana/metrictank/blob/master/expr/NOTES#L35

expr/func_aspercent.go
```diff
@@ -0,0 +1,241 @@
...
+		serie.Target = fmt.Sprintf("asPercent(%s,%s)", serie.Target, totalsSerie.QueryPatt)
+		serie.Tags = map[string]string{"name": serie.Target}
+		for i := range serie.Datapoints {
+			var totalVal float64
+			if len(totalsSerie.Datapoints) > i {
+				totalVal = totalsSerie.Datapoints[i].Val
+			} else {
+				totalVal = s.totalFloat
+			}
+			serie.Datapoints[i].Val = computeAsPercent(serie.Datapoints[i].Val, totalVal)
```
shanson7: I think this violates https://github.com/grafana/metrictank/blob/master/expr/NOTES#L35

expr/func_aspercent.go
stivenbb: very nice, I like it

expr/func_aspercent.go
stivenbb: fixed all and added a test

expr/func_aspercent_test.go
```diff
@@ -555,6 +563,28 @@ func testAsPercent(name string, in []models.Series, out []models.Series, t *test
 			t.Fatalf("case %q (%q, %v): output point %d - expected %v got %v", name, total, nodes, j, p, g.Datapoints[j])
 		}
 	}
+
+	// Test if original series was modified
+	for i, orig := range originalSeries {
+		inSerie := in[i]
+		if orig.Target != inSerie.Target {
```
shanson7: Could you use something like https://github.com/grafana/metrictank/blob/0428430b501132244d3a4e94ba8e280c640b07b9/expr/plan_test.go#L186 here?

expr/func_aspercent_test.go
```diff
@@ -526,6 +526,14 @@ func testAsPercent(name string, in []models.Series, out []models.Series, t *test
 	}
 	f.(*FuncAsPercent).totalFloat = totalFloat
 	f.(*FuncAsPercent).nodes = nodes
+
+	originalSeries := make([]models.Series, len(in))
+	for i, serie := range in {
+		originalSeries[i].Interval = serie.Interval
+		originalSeries[i].QueryPatt = serie.QueryPatt
+		originalSeries[i].Target = serie.Target
+		originalSeries[i].Datapoints = getCopy(serie.Datapoints)
```
shanson7: maybe just
```
originalSeries[i] = serie
originalSeries[i].Datapoints = getCopy(serie.Datapoints)
```
Gets all the things like tags etc.

expr/func_aspercent.go
```diff
@@ -130,6 +130,8 @@ func (s *FuncAsPercent) Exec(cache map[Req][]models.Series) ([]models.Series, er
 		if len(totals) == 1 {
 			totalsSerie = totals[0]
 		} else if len(totals) == len(series) {
+			// Sorted to match the input series with the total series based on Target.
+			// Mimicks Graphite's implementation
```
shanson7: Nit, but 'Mimic' the common modern spelling

expr/func_aspercent.go
```diff
@@ -239,3 +243,11 @@ Loop:
 		Tags:         map[string]string{"name": name},
 	}
 }
+
+func copyDatapoints(serie *models.Series) {
+	out := pointSlicePool.Get().([]schema.Point)
+	for _, p := range serie.Datapoints {
```
shanson7:
```
out = append(out, serie.Datapoints...)
```

expr/func_aspercent_test.go
stivenbb: No, because the original series does get modified by metrictank (a "name" tag gets added). Makes sense to compare relevant values only.

stivenbb:
Note: this requires Go version 1.10+ because it uses math.Round()

@shanson7 any outstanding comments?

stivenbb:
Any chance this can get merged to master?

Dieterbe:
do you plan to make more changes or dyou consider this complete?
also, it's kindof hard to see which code came from my old PR, is it possible to clarify the situation by reusing some of those commits. ideally perhaps, this PR could resume where the other branch left off.
but if all of this is too much work, then never mind.

stivenbb:
This is good to go. I just rebased it to make the merge cleaner.

And I didn't use much, just added an additional argument type. It would be a lot of work to reuse the old commits with little to no payoff imo, since I rewrote most of it.

expr/plan.go
```diff
@@ -198,15 +198,18 @@ func newplanFunc(e *expr, fn GraphiteFunc, context Context, stable bool, reqs []
 	// this function, we can set up the input arguments for the function
 	// that are series
 	pos = 0
-	for _, argExp = range argsExp[:cutoff] {
+	for _, argExp = range argsExp {
+		if pos >= len(e.args) {
+			break // no more args specified. we're done.
+		}
```
Dieterbe: the removal of `cutoff` here is this safe?

expr/func_aspercent.go
Dieterbe: defer args are evaluated at call time, so this won't work as expected.
see https://play.golang.org/p/06o0zrVeOH-

stivenbb: You forgot to call foo() in your snippet.
see https://play.golang.org/p/Qh4rp1p2q73

expr/func_aspercent.go
```diff
@@ -0,0 +1,258 @@
...
+
+		// calculate the sum
+		if math.IsNaN(s.totalFloat) && totals == nil {
+			totalSeries = getTotalSeries(metaSeries)
+			// calculate sum of totals series
+		} else if totals != nil {
+			totalSeriesLists := groupSeriesByKey(totals, s.nodes, &keys)
+			totalSeries = getTotalSeries(totalSeriesLists)
+		} else {
+			return nil, errors.New("total must be None or a seriesList")
```
Dieterbe: seems like it's fairly easy to trigger this by specifying a serieslist pattern that doesn't match any series.
maybe in that case we can provide a better error message? (basically above at `if len(totals) == 0 {` maybe directly return an error ? what does graphite do in this case?

expr/func_aspercent.go
```diff
@@ -0,0 +1,258 @@
...
+		},
+		ArgStringsOrInts{val: &s.nodes, opt: true, key: "nodes"},
+	}, []Arg{ArgSeriesList{}}
+}
+
+func (s *FuncAsPercent) Context(context Context) Context {
+	return context
+}
+
+func (s *FuncAsPercent) Exec(cache map[Req][]models.Series) ([]models.Series, error) {
```
Dieterbe: this function is rather complex. can we split it up in smaller functions?

expr/func_aspercent.go
stivenbb: I'm not sure what you mean here. This error message gets triggered if they pass in a number. In the case where they pass in a nodes argument, only a series or nothing should be passed in for the `total` argument. First branch is triggered if neither is passed in. Second branch is triggered if a series is passed in. Which leaves the case where a number is passed in (i.e `math.IsNaN(s.totalFloat) = false`)

Graphite throws that exact error message is that case.

Dieterbe: if you specify a serieslist pattern that doesn't match any series, then I believe this will happen:
```
        if s.totalSeries != nil {
                totals, err = s.totalSeries.Exec(cache)
                if err != nil {
                        return nil, err
                }
                if len(totals) == 0 {
                        totals = nil <---- this right here
                }
        }
```

expr/plan.go
Dieterbe: looks good :) the argument consumption/iterating code is probably my least favorite code of MT. FWIW.
if you can handle this, you can handle anything else :dancing_men:

expr/func_aspercent.go
```diff
@@ -0,0 +1,268 @@
...
+
+	if s.nodes != nil {
+		outSeries, err = s.execWithNodes(series, totals)
+	} else {
+		outSeries, err = s.execWithoutNodes(series, totals)
+	}
+	return outSeries, err
+}
+
+func (s *FuncAsPercent) execWithNodes(series []models.Series, totals []models.Series) ([]models.Series, error) {
```
Dieterbe: `series, totals []models.Series`. also the other function

expr/func_aspercent.go
```diff
@@ -0,0 +1,268 @@
...
+		QueryPatt:    name,
+		Datapoints:   out,
+		Interval:     series[0].Interval,
+		Consolidator: cons,
+		QueryCons:    queryCons,
+		Tags:         map[string]string{"name": name},
+	}
+}
+
+func deepCopySerieElements(serie *models.Series) {
```
Dieterbe: this API is a bit strange. would it not make more sense to return a new series that is a deep copy?
even though that is slightly more work, seems worth it.

in fact, maybe add a `Copy()` method to the serie type in the models package

Dieterbe: also looking at sumSeries, the copying/newly-allocating seems a bit too eager.
we should only need to allocate point slices:
1) for the totals slice: when we need to sum up values and we need a place to store the totals (not when totals results in a single series, then we can just read from it)
2) for each output series.(if it is different from the input slice)

Dieterbe:
some unit tests for ArgIn would be nice.
with some luck you can just cherry pick 47ff80ac2898ae92083e4be1f52c22ebfc2e519d

expr/func_aspercent.go
stivenbb: FWIW, Graphite crashes if you do that.

Let me see if I can handle it in some clean way

expr/func_aspercent.go
stivenbb: The problem with #1 is that later I modify the total series in some cases such as this (line 99):

```
// No series for total series
if _, ok := metaSeries[key]; !ok {
	serie2 := totalSeries[key]
	serie2.QueryPatt = fmt.Sprintf("asPercent(MISSING,%s)", serie2.QueryPatt)
	serie2.Target = fmt.Sprintf("asPercent(MISSING,%s)", serie2.Target)
	serie2.Tags = map[string]string{"name": serie2.Target}
	for i := range serie2.Datapoints {
		serie2.Datapoints[i].Val = math.NaN()
	}
	outSeries = append(outSeries, serie2)
	continue
}
```

stivenbb:
Found a bug. ArgIn doesn't support keyword args because as of now series keyword args aren't supported. I will refactor a bit to make it not crash, but adding support for series through keyword args should be a goal for the future.

expr/func_aspercent.go
stivenbb: Ok, so this was just unnecessary:
```
if len(totals) == 0 {
            totals = nil
}
```

So I removed it. Now if a series returns empty, it does not assume that there were no arguments

expr/func_aspercent.go
Dieterbe: I'll be mostly afk until Thursday.
Please read expr/NOTES carefully if you have not already done so. All this complexity is to make sure we never overwrite data in the memory AggMetric, chunk cache etc.
Some of these fields are by value so harmless but the tags map is interesting as well. Not sure if we currently take that into account everywhere. On phone so can't check right now

expr/func_aspercent_test.go
stivenbb: Scratch that. the real problem was that `reflect.DeepEqual(math.NaN(), math.NaN()) == false`, which is not the case when comparing series

expr/func_aspercent.go
stivenbb: Ok, well I added a `Series.Copy(emptyDatapoints []schema.Point)` function (the argument is there so that `pointSlicePool` can be used if needed). And I only copy the series if I modify values that shouldn't be modified.

Hopefully that's sufficient and what you were looking for 😃

Dieterbe:
I pushed a few extra commits to https://github.com/grafana/metrictank/tree/asPercent2
please check it out. ideally all 3 can be merged into your branch. the first two are more cosmetic,
the third introduces tests and a utility function which i had in my original branch.
these tests are proving to be useful:

```
--- FAIL: TestAsPercentSingleInputUsingSelf (0.00s)
	func_aspercent_old_test.go:129: series 0: Target "asPercent(a)", got "asPercent(a,sumSeries(a))"
--- FAIL: TestAsPercentDoubleInputUsingSelf (0.00s)
	func_aspercent_old_test.go:165: series 0: Target "asPercent(a.a)", got "asPercent(a.a,sumSeries(a.*))"
--- FAIL: TestAsPercentDoubleInputUsingOne (0.00s)
	func_aspercent_old_test.go:211: series 0: Target "asPercent(a.a,foo.a)", got "asPercent(a.a,foo.*)"
--- FAIL: TestAsPercentSingleInputUsingFloat10 (0.00s)
	func_aspercent_old_test.go:333: series 0: point 1 - expected {506 20} got {506.00000000000006 20}
FAIL
exit status 1
FAIL	github.com/grafana/metrictank/expr	0.090s
```
the first 2 definitely seem legit. the 3rd i'm not able to confirm via play.grafana.org as that seems to be an older graphite version which doesn't support multiple series for `total` (if you have a graphite server would you mind checking?)
the last test case could be considered bogus but it depends on how exactly you compute the results, maybe an inaccuracy is introduced somewhere.

I think the style of calling equalOutput in each test function is good, so that if a test fails, you can immediately jump to the right location (e.g. in vim's quick fix menu and i'm sure other editors have an equivalent), otherwise all failures just end up in testAsPercent

stivenbb:
Can you explain the difference between QueryPatt and Target? Like theoretically I get what it's trying to do, but I feel like they are very inconsistent between functions/compared to graphite.

Dieterbe:
> difference between QueryPatt and Target

the former is the pattern like `foo.*`, the latter is typically more concrete like `foo.bar`. sometimes they are the same, (e.g. when query pattern is a concrete identifier without wildcards).
when in doubt, do whatever graphite does (we generally mimic it pretty well though probably not 100%, but it's the goal.

expr/func_aspercent.go
```diff
@@ -0,0 +1,251 @@
...
+
+	var outSeries []models.Series
+	defer func() { cache[Req{}] = append(cache[Req{}], outSeries...) }()
+	var totals []models.Series
+	if s.totalSeries != nil {
+		totals, err = s.totalSeries.Exec(cache)
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
stivenbb: If totals is zero length, it is not an error. Instead, it will return a bunch of empty series with names something like `asPercent(foo.*,MISSING)`

expr/func_aspercent.go
```diff
@@ -0,0 +1,251 @@
...
+				serie1.Target = fmt.Sprintf("asPercent(%s,%s)", serie1.Target, serie2.Target)
+				serie1.Tags = map[string]string{"name": serie1.Target}
+				for i := range serie1.Datapoints {
+					serie1.Datapoints[i].Val = computeAsPercent(serie1.Datapoints[i].Val, serie2.Datapoints[i].Val)
+				}
+				outSeries = append(outSeries, serie1)
+			}
+			return outSeries, nil
+		} else {
+			return nil, errors.New("asPercent second argument must be missing, a single digit, reference exactly 1 series or reference the same number of series as the first argument")
```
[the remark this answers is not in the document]
stivenbb: This is just mimicking Graphite. Should I still say it?

expr/func_aspercent.go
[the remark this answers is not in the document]
stivenbb: Ohhh I see what you mean. Because the totalsSeries has to be either `== 1` or `== len(series)`

But the error will get triggered at a later check. Plus, if the series is empty as well, then it should trigger an error technically

expr/func_aspercent.go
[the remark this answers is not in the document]
Dieterbe: Ha ! Yeah for errors I think deviating if we can be more clear is fine . The replicating exact behavior is more important for functionality, names of returned series, etc .

expr/func_aspercent.go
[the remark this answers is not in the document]
Dieterbe: Generally I think it's better to return the error early when we can. So when reasoning about the code we have fewer special cases that complicate things further down.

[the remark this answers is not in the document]
stivenbb: Ok, I moved up the validation like you suggested, so the check for proper length of totals is right after it. (Gonna push soon just looking at one more thing)

stivenbb:
So target naming must've been different in older versions of Graphite because I'm using the latest one, and the Targets match perfectly for all use cases. That being said, sometimes Target and QuerryPatt are the same, so it can be hard to see which one to use.
For `TestAsPercentSingleInputUsingSelf` and `TestAsPercentDoubleInputUsingSelf`, the graphite code for setting targets is:
https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L905
and
https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L930
You can clearly see that it sets ` name = "asPercent(%s,%s)" % (series.name, totalText or series.pathExpression)`, not just the `series.name` like you have in your test cases.

`TestAsPercentDoubleInputUsingOne` did catch an inconsistency, which I fixed.

As for `TestAsPercentSingleInputUsingFloat10`, seems to be a weird rounding thing with Go: https://play.golang.org/p/3geMM83UH0f
I just changed the test case to use float64()

Also, in `consumeSeriesArg` I added a check to ignore arguments that are equal to `None`. That way, it will interpret `None` as `nil`

stivenbb:
For my change in `consumeSeriesArg`, I'm not sure if what I did is sufficient. `None` should only be acceptable for optional arguments, so I need to add a check somewhere to throw an error if None is passed for non-optional arguments. Will push a change shortly.

shanson7:
@Dieterbe - Is it ok that this change bumps the minimum supported compiler version to 1.10?

Dieterbe:
That's fine. Please grep around to see if we document go version somewhere and update accordingly

Dieterbe:
re QueryPattern vs Target, see also https://godoc.org/github.com/grafana/metrictank/api/models#Series for more info.

> You can clearly see that it sets name = "asPercent(%s,%s)" % (series.name, totalText or series.pathExpression), not just the series.name like you have in your test cases.

interesting. perhaps a recent change in graphite.
I can confirm using latest official graphite docker image, a query `asPercent(carbon.agents.graphitemon-a.cache.queues)` results in : 
```
"target": "asPercent(carbon.agents.graphitemon-a.cache.queues,sumSeries(carbon.agents.graphitemon-a.cache.queues))",
"tags": {"name": "asPercent(carbon.agents.graphitemon-a.cache.queues,sumSeries(carbon.agents.graphitemon-a.cache.queues))"}
```

expr/func_aspercent_old_test.go
```diff
@@ -28,7 +28,7 @@ var a1 = []schema.Point{
 var b1 = []schema.Point{
 	{Val: 10, Ts: 10},
 	{Val: -25.8, Ts: 20},
-	{Val: math.MaxFloat64 / 1000, Ts: 30},
+	{Val: float64(math.MaxFloat64 / 1000), Ts: 30},
```
Dieterbe: what's the goal of adding `float64(..)` here? .Val is already typed float64

expr/func_aspercent_old_test.go
```diff
@@ -53,8 +53,8 @@ var a1AsPercentOfa1 = []schema.Point{
 }
 var a1AsPercentOfa1b1 = []schema.Point{
 	{Val: 0, Ts: 10}, // 100 * 0 / (0+10)
-	{Val: 100 * float64(50.6) / float64(50.6-25.8), Ts: 20},
-	{Val: 100 * float64(1234567890) / (float64(1234567890) + math.MaxFloat64/1000), Ts: 30},
+	{Val: float64(50.6) / float64(50.6-25.8) * 100, Ts: 20},
+	{Val: float64(1234567890) / (float64(1234567890) + math.MaxFloat64/1000) * 100, Ts: 30},
```
Dieterbe: these changes make sense. to first divide, then multiply by 100.
multiplying by 100 first as I had it, could lead to overflows

expr/func_aspercent_old_test.go
```diff
@@ -77,9 +77,9 @@ var b1AsPercentOfa1b1b1 = []schema.Point{
 }
 
 var b1AsPercentOfa1 = []schema.Point{
-	{Val: math.MaxFloat64, Ts: 10},
-	{Val: 100 * (-25.8) / 50.6, Ts: 20},
-	{Val: 100 * (math.MaxFloat64 / float64(1000)) / float64(1234567890), Ts: 30},
+	{Val: math.NaN(), Ts: 10},
```
Dieterbe: i can confirm. in graphite. 100/0 = None. technically, should be inf, but graphite has no concept for inf

expr/func_aspercent_test.go
```diff
@@ -0,0 +1,649 @@
...
+		nil,
+		"123.456",
+	)
+}
+
+func TestAsPercentTotalSerie(t *testing.T) {
+	out1 := []schema.Point{
+		{Val: math.NaN(), Ts: 10},
+		{Val: math.NaN(), Ts: 20},
+		{Val: math.Inf(0), Ts: 30},
```
Dieterbe: hmm? b/a = `(math.MaxFloat64 - 20) / 5.5`, what am i missing here?
anyway as i suggested yesterday I think the tests should be rewritten so that the failures are within the Test functions and not within the testAsPercent helper.

expr/func_aspercent_test.go
```diff
@@ -0,0 +1,649 @@
...
+		{Val: math.NaN(), Ts: 50},
+		{Val: math.NaN(), Ts: 60},
+	}
+	out2 := []schema.Point{
+		{Val: math.NaN(), Ts: 10},
+		{Val: math.NaN(), Ts: 20},
+		{Val: float64(199) * 100, Ts: 30},
+		{Val: float64(29) / 2 * 100, Ts: 40},
+		{Val: float64(80) / 3.0 * 100, Ts: 50},
+		{Val: float64(250) / 4 * 100, Ts: 60},
```
Dieterbe: shouldn't out2 be the same as in the previous function? because both cases have `asPercent(d,a)`

expr/func_aspercent.go
```diff
@@ -0,0 +1,254 @@
...
+		serie.QueryPatt = fmt.Sprintf("asPercent(%s,%s)", serie.QueryPatt, totalsSerie.QueryPatt)
+		serie.Target = fmt.Sprintf("asPercent(%s,%s)", serie.Target, totalsSerie.Target)
+		serie.Tags = map[string]string{"name": serie.Target}
+		for i := range serie.Datapoints {
+			var totalVal float64
+			if len(totalsSerie.Datapoints) > i {
+				totalVal = totalsSerie.Datapoints[i].Val
+			} else {
+				totalVal = s.totalFloat
+			}
```
Dieterbe: isn't totalFloat pretty much guaranteed to be NaN here?

expr/func_aspercent.go
```diff
@@ -0,0 +1,254 @@
...
+
+	for _, serie := range series {
+		serie = serie.Copy(pointSlicePool.Get().([]schema.Point))
+		serie.QueryPatt = fmt.Sprintf("asPercent(%s,%s)", serie.QueryPatt, totalsSerie.QueryPatt)
+		serie.Target = fmt.Sprintf("asPercent(%s,%s)", serie.Target, totalsSerie.Target)
+		serie.Tags = map[string]string{"name": serie.Target}
+		for i := range serie.Datapoints {
+			var totalVal float64
+			if len(totalsSerie.Datapoints) > i {
+				totalVal = totalsSerie.Datapoints[i].Val
```
Dieterbe: we may want to always just assume this branch is taken and just write this code line directly without the if/else.
that way if we ever have a bug where `len(totalSerie.Datapoints) != len(serie.Datapoints)` we can troubleshoot it instead of trying to hide such error case and returning incorrect data.

expr/func_aspercent.go
```diff
@@ -0,0 +1,254 @@
...
+		} else if len(totals) == len(series) {
+			// Sorted to match the input series with the total series based on Target.
+			// Mimicks Graphite's implementation
+			sort.Slice(series, func(i, j int) bool {
+				return series[i].Target < series[j].Target
+			})
+			sort.Slice(totals, func(i, j int) bool {
+				return totals[i].Target < totals[j].Target
+			})
+			for i, serie1 := range series {
```
Dieterbe: lots of duplication here wrt the similar code block further down.
can't we do all the if/else stuff here to just set up the right totals and series variables,
and then do the same processing at the end irrespective of which scenario it was?

expr/func_aspercent.go
```diff
@@ -0,0 +1,254 @@
...
+	var totalSeries map[string]models.Series
+
+	// calculate the sum
+	if math.IsNaN(s.totalFloat) && totals == nil {
+		totalSeries = getTotalSeries(metaSeries)
+		// calculate sum of totals series
+	} else if totals != nil {
+		totalSeriesLists := groupSeriesByKey(totals, s.nodes, &keys)
+		totalSeries = getTotalSeries(totalSeriesLists)
+	}
```
Dieterbe: * we compute totals even for keys we won't need (e.g. keys not in the input series)
* for each missing case we repeatedly get a slice and fill it with NaN's. we could reuse the same slice in this case. my suggestion would be declare a `var nones []schema.Point`. then whenever you need it, if it's nil, instantiate it. if it's not, just reuse it.

Dieterbe:
hi @stivenbb my first pass of review comments is now done :p let me know when you've addressed everything or if you have any questions.

expr/func_aspercent.go
shanson7: I don't think that would play nicely with the pointSlicePool, would it?

Dieterbe: oh right, because at the end the same slice would get added to the pool multiple times, which is bad. so nevermind that then.

Dieterbe: On phone now
best would be to just add the slice to the cache thing once I think

expr/func_aspercent_test.go
stivenbb: * 100 -> overflow

expr/func_aspercent_test.go
stivenbb: No, because when totals is a seriesList with `len(totals) == len(series)`, the series first get sorted by tag before getting matched. So, in this test case it would be asPercent(d,c) and asPercent(b,a). This behavior is same in Graphite

expr/func_aspercent.go
stivenbb: No, because totalFloat is still a valid option here. (the `else` part of the if statement right before this)

expr/func_aspercent.go
stivenbb: this is more of a check of whether there is a totalSeries at all. I'll change it to `> 0`, that way we get an index out of range exception if there's a bug.

stivenbb:
@Dieterbe ok, I think I've addressed all the requested changes. Let me know if I missed something

Dieterbe:
> what is it that requires the new go version?

expr/func_aspercent.go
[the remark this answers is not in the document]
Dieterbe: @stivenbb ^^ why can't we just return an error here if `len(totals)` is 0 ? this case still slips through into the `if s.nodes != ni` case for example

shanson7:
https://github.com/grafana/metrictank/pull/966#issuecomment-408935104

math.Round

expr/func_aspercent.go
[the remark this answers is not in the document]
stivenbb: When `nodes != nil` and `len(totals) == 0`, it shouldn't fail. Instead, all targets will be `asPercent(foo,MISSING)`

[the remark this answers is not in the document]
Dieterbe: aha! ok thx

Dieterbe:
I think the last things to do here are : 
* I added 3 commits to https://github.com/grafana/metrictank/tree/asPercent2 , please review them and merge / cherry-pick them into your branch if they look good to you.
* in execWithNodes we still compute totals for all aggKeys, even for keys not found in the metaSeries. this is needless extra work

stivenbb:
Ok, just cherry picked. Will work on not computing totals when unnecessary.

Dieterbe:
oh i realized my code `nones := pointSlicePool.Get().([]schema.Point)` is wrong, should be `=` to prevent scoping issues I think.

stivenbb:
cassandra test seems to be failing after I cherry-picked... Do you know what that could be about?

stivenbb:
Ok, should be good to go now. LMK if my change looks good.

Dieterbe:
great work @stivenbb, thank you very much for your work on this.

PR #1103, opened 2018-10-23
set sarama client KafkaVersion via config

woodsaj:
see issue: #1053

from  https://github.com/Shopify/sarama/blob/v1.19.0/config.go#L324-L330

```
The version of Kafka that Sarama will assume it is running against.
Defaults to the oldest supported stable version. Since Kafka provides
backwards-compatibility, setting it to a version older than you have
will not break anything, although it may prevent you from using the
latest features. Setting it to a version greater than you are actually
running may lead to random breakage.
```

### DIFF
--- a/docker/docker-chaos/metrictank.ini
+++ b/docker/docker-chaos/metrictank.ini
@@ -224,8 +224,8 @@ enabled = true
 org-id = 0
 # tcp address (may be given multiple times as a comma-separated list)
 brokers = kafka:9092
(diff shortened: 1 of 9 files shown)


woodsaj:
> seems like sarama supports any delimiter between the digits

this does appear to be true, but good catch. i just assumed that passing the supportVersion names would work, but ParseKafkaVersion actually wants the semver string. eg. `0.10.0.0`

woodsaj:
@Dieterbe  this should be good now

PR #1187, opened 2018-12-25
debug build of metrictank and debug docker/k8s images

robert-milan:
add docker-dev-debug for running local metrictank with dlv in container
add -debug flag for use with build scripts
add Dockerfiles for local and kubernetes versions
add .vscode to .gitignore
add metrictank-debug image for public and private repositories


Makefile
```diff
@@ -26,11 +25,35 @@ qa: bin
 	scripts/qa/vet-high-confidence.sh
 	# qa-post-build steps minus stack tests
 	scripts/qa/docs.sh
+
+#debug versions for remote debugging with delve
+bin-debug:
+	./scripts/build.sh -debug
+docker-debug:
+	./scripts/build_docker.sh -debug
+qa-debug: bin-debug
```
Dieterbe: let's isolate out all the qa commands so we don't have the repetition. (e.g. into a qa-common target)
you can then make qa-debug and qa depend on qa-common

docker/docker-dev-debug/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
+version: '2'
+
+services:
+  metrictank:
+    security_opt:
+      - seccomp:unconfined
```
Dieterbe: do you have a link that best describes why we need this?
i googled, but mostly found random support tickets.

docker/docker-dev-debug/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+    hostname: metrictank
+    image: grafana/metrictank:latest-debug
+    ports:
+     - "6060:6060"
+     - "2003:2003"
+     - "40000:40000"
+    volumes:
+      - ../../build/metrictank:/usr/bin/metrictank
+      - ./metrictank.ini:/etc/metrictank/metrictank.ini
+      - ./storage-schemas.conf:/etc/metrictank/storage-schemas.conf
```
Dieterbe: do we need the custom storage-schemas.conf and storage-aggregation.conf?

docker/docker-dev-debug/metrictank.ini
```diff
@@ -0,0 +1,417 @@
+## misc ##
```
Dieterbe: this file is out of sync with `../../scripts/config/metrictank-docker.ini`. seems the only change we should have is the `prefix` setting

Dieterbe:
need to update docs/docker.md

scripts/build.sh
```diff
@@ -38,6 +38,11 @@ for bin in *; do
     set -x
     # -race requires CGO
     CGO_ENABLED=1 go build -race -ldflags "-X main.version=$version" -o $BUILDDIR/$bin || fail
+  elif [ "$1" == "-debug" ]
+  then
+    set -x
+    # -debug flags
+    CGO_ENABLED=0 go build -gcflags "all=-N -l" -ldflags "-X main.version=${version}-debug" -o $BUILDDIR/$bin || fail
```
Dieterbe: are race detector and dlv mutually exclusive?
seems both should work at the same time.
if so, perhaps we should always build the debug images?

scripts/dlv/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN mkdir -p /etc/metrictank /usr/share/metrictank/examples
+COPY config/metrictank-docker.ini /etc/metrictank/metrictank.ini
+COPY config/index-rules.conf /etc/metrictank/index-rules.conf
+COPY config/storage-schemas.conf /etc/metrictank/storage-schemas.conf
+COPY config/storage-aggregation.conf /etc/metrictank/storage-aggregation.conf
+COPY config/schema-store-cassandra.toml /etc/metrictank/schema-store-cassandra.toml
+COPY config/schema-idx-cassandra.toml /etc/metrictank/schema-idx-cassandra.toml
+COPY config/schema-store-scylladb.toml /usr/share/metrictank/examples/schema-store-scylladb.toml
+COPY config/schema-idx-scylladb.toml /usr/share/metrictank/examples/schema-idx-scylladb.toml
```
Dieterbe: can't we do this without repeating all the stuff we did for the regular image?
maybe use `FROM metrictank` ? (obviously, requires that image to be built before building this one)

docker/docker-dev-debug/docker-compose.yml
robert-milan: I will add this as a comment to the docker-compose file as well
https://github.com/Microsoft/vscode-go/wiki/Debugging-Go-code-using-VS-Code#linuxdocker

docker/docker-dev-debug/docker-compose.yml
robert-milan: I suppose we don't need custom storage-schemas or aggregation configuration files. I will remove them.

docker/docker-dev-debug/metrictank.ini
robert-milan: Yes. I believe I may have initially copied the wrong file when I was putting this together, I will fix that.

scripts/build.sh
Dieterbe: > and only one binary is produced from build.sh 
this is incidental. not by design.
everything else you say here seems fair enough.

scripts/k8s_dlv/entrypoint_debug.sh
```diff
@@ -0,0 +1,86 @@
...
+  pkill metrictank
+  sleep 1
+  echo "Killing dlv"
+  pkill dlv
+  sleep 1
+  exit 0
+}
+
+trap 'kill_metrictank' SIGTERM
+trap 'kill_metrictank' SIGINT
```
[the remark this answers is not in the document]
robert-milan: Yes it doesn't properly terminate Metrictank and then the container hangs. I will document it in a comment.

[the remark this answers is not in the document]
Dieterbe: it's unfortunate, all the complexity we have to add to deal with this..
can we solve this in a more clean way, e.g. add support to delve to properly kill its child process when it is terminated? or is it a design goal to keep the child alive?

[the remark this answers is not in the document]
robert-milan: I'm looking into other possible options or flags to use that might accomplish this.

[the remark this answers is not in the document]
robert-milan: I haven't been able to find another way to get this done.

Makefile
```diff
@@ -26,11 +34,17 @@ qa: bin
...
+
 all:
 	$(MAKE) bin
 	$(MAKE) docker
 	$(MAKE) qa
 
+debug:
+	$(MAKE) bin-debug
+	$(MAKE) docker-debug
+	${MAKE} qa-debug
```
replay: the last `MAKE` has `{}` instead of `()`. is that intentional?

scripts/build.sh
replay: just for consistency with `:40` i'd use `$version` instead of `${version}`.

scripts/util/wait_for_endpoint_debug.sh
```diff
@@ -0,0 +1,91 @@
...
+# when it receives SIGTERM (docker stop / docker-compose stop / etc...). So, instead we leave
+# this script running in an endless sleep loop and trap SIGTERM, SIGINT, and SIGHUP. Then we can
+# kill both metrictank and dlv and exit this script which will shutdown the container.
+kill_metrictank() {
+  echo "Killing metrictank"
+  pkill metrictank
+  sleep 1
+  echo "Killing dlv"
+  pkill dlv
+  sleep 1
```
replay: what's the purpose of the second sleep? i assume the first one is because you want to wait with killing dlv until mt has exited?

Makefile
robert-milan: No, that's a mistake

scripts/util/wait_for_endpoint_debug.sh
robert-milan: I don't recall currently. Do you think there is any harm leaving it in?

replay: nope, no harm

Dieterbe:
@robert-milan is this ready to merge in your opinion?

robert-milan:
> @robert-milan is this ready to merge in your opinion?

I believe so, I would like to have a day or two more just to look over it again.

Issue #1327, opened 2019-05-24
Package level stats with Partitioned Index


robert-milan:
We currently use a lot of package level and project global stats in Metrictank. This has worked fine in the past because we have been using the Unpartitioned Index. When we switch over to using the Partitioned Index more often I think this will be a problem and stats will be reported / recorded incorrectly. There will more than likely be race conditions in reporting as well.

There are multiple solutions to the problem. It really depends on the nature of the stat. If we are dealing with a counter that we are adding to and subtracting from then it should already work, but if we are looking at a single number that is updated, then it will always reflect the latest Unpartitioned Index that updated it. Instead, we may need to do some kind of moving average internally. I think that is part of a larger discussion on how we want to report our own internal stats.

woodsaj:
The only issue with the stats in the idx.Memory package is the use of gauges where the value is "set". 
 But the only gauge where this is done is`statMetricsActive` .  `Set()` is used 3 times (in deleteTaggedByIdSet(), Delete() and Prune()) and these can be easily changed to just call `Dec()` instead.


All other stats are counters or histograms and so work across multiple partitions.

As a priority we should fix `statMetricsActive`.   

However, i would also prefer to move away from package level stats, but that is less of a priority

robert-milan:
`statsMetricsActive` should now be fixed. I'll look at another PR for the rest of them in a few weeks.

PR #1336, opened 2019-05-29
Fix active series stats decrement

robert-milan:
Fixes issue while decrementing certain memory stats from a partitioned index. This does not close the issue, but addresses the most urgent problem.

See also: #1327


shanson7:
Take a look at #1169 and #1132 to prevent re-introducing the issue. It's possible that the underlying problem was addressed in #1303

robert-milan:
I believe that https://github.com/grafana/metrictank/pull/1336/files#diff-051eec914f9c2892a0d9f0d2a4d87336R1568 addresses the previous issue of deleting more than it was supposed to. I don't think this was being accounted for in earlier iterations.

PR #1370, opened 2019-06-26
WIP: add concurrency limits on find requests

woodsaj:
- Adds a max-concurrent-finds config option and code to limit
  the number of Finds() that can be running at any one time.
- Find() and findMaybeCached() functions have been refactored
  to avoid concurrency limits being applied to queries that
  are served from the findCache.


woodsaj:
To make the concurrency limits useful, we really need to have find requests that are blocked for too long fail.  This should be achieved by passing a context through to the request as mentioned in issue #1362

PR #1382, opened 2019-07-05
New flag 'ingest-from'

fkaleo:
New flag 'ingest-from' that instructs metrictank to discard any point of a specific org id that does not belong to chunks starting after a given timestamp.
e.g. 20:67031 with a chunk span of 10 will discard all points with org id 20 and timestamps lesser than 67040

### DIFF
--- a/cmd/metrictank/metrictank.go
+++ b/cmd/metrictank/metrictank.go
@@ -298,14 +298,12 @@ func main() {
 		Initialize our MemoryStore
 	***********************************/
 
-	ingestFrom := make(map[uint32]int64)
-	ingestFromStrPerOrg := strings.Split(*ingestFromStr, ",")
-	for _, ingestFromStrForOrg := range ingestFromStrPerOrg {
-		ingestFromOrgID, ingestFromTimestamp := util.MustParseIngestFromFlag(ingestFromStrForOrg)
-		if ingestFromTimestamp > 0 {
-			log.Infof("For org %d, will only ingest data for chunks that have a t0 equal or higher to %s", ingestFromOrgID, time.Unix(ingestFromTimestamp, 0))
-		}
-		ingestFrom[ingestFromOrgID] = ingestFromTimestamp
+	ingestFrom, err := util.ParseIngestFromFlags(*ingestFromStr)
+	if err != nil {
+		log.Fatalf("ingest-from: %s", err.Error())
+	}
+	for orgID, timestamp := range ingestFrom {
+		log.Infof("For org %d, will only ingest data for chunks that have a t0 equal or higher to %s", orgID, time.Unix(timestamp, 0))
 	}
 	if inputEnabled {
 		metrics = mdata.NewAggMetrics(store, ccache, *dropFirstChunk, ingestFrom, chunkMaxStale, metricMaxStale, gcInterval)
--- a/cmd/metrictank/metrictank.go
+++ b/cmd/metrictank/metrictank.go
@@ -302,8 +302,10 @@ func main() {
 	if ingestAfterTimestamp > 0 {
 		log.Infof("Will only ingest data points for org id %d belonging to chunks starting after %s", ingestAfterOrgID, time.Unix(ingestAfterTimestamp, 0))
 	}
+	ingestAfter := make(map[uint32]int64)
+	ingestAfter[ingestAfterOrgID] = ingestAfterTimestamp
 	if inputEnabled {
-		metrics = mdata.NewAggMetrics(store, ccache, *dropFirstChunk, ingestAfterOrgID, ingestAfterTimestamp, chunkMaxStale, metricMaxStale, gcInterval)
+		metrics = mdata.NewAggMetrics(store, ccache, *dropFirstChunk, ingestAfter, chunkMaxStale, metricMaxStale, gcInterval)
 	}
 
 	/***********************************
--- a/cmd/metrictank/metrictank.go
+++ b/cmd/metrictank/metrictank.go
@@ -298,12 +298,15 @@ func main() {
 		Initialize our MemoryStore
 	***********************************/
 
-	ingestAfterOrgID, ingestAfterTimestamp := util.MustParseIngestAfterFlag(*ingestAfterStr)
-	if ingestAfterTimestamp > 0 {
-		log.Infof("Will only ingest data points for org id %d belonging to chunks starting after %s", ingestAfterOrgID, time.Unix(ingestAfterTimestamp, 0))
-	}
 	ingestAfter := make(map[uint32]int64)
-	ingestAfter[ingestAfterOrgID] = ingestAfterTimestamp
+	ingestAfterStrPerOrg := strings.Split(*ingestAfterStr, ",")
+	for _, ingestAfterStrForOrg := range ingestAfterStrPerOrg {
+		ingestAfterOrgID, ingestAfterTimestamp := util.MustParseIngestAfterFlag(ingestAfterStrForOrg)
+		if ingestAfterTimestamp > 0 {
+			log.Infof("Will only ingest data points for org id %d belonging to chunks starting after %s", ingestAfterOrgID, time.Unix(ingestAfterTimestamp, 0))
+		}
+		ingestAfter[ingestAfterOrgID] = ingestAfterTimestamp
+	}
 	if inputEnabled {
 		metrics = mdata.NewAggMetrics(store, ccache, *dropFirstChunk, ingestAfter, chunkMaxStale, metricMaxStale, gcInterval)
 	}
--- a/cmd/metrictank/metrictank.go
+++ b/cmd/metrictank/metrictank.go
@@ -60,7 +60,7 @@ var (
 
 	// Data:
 	dropFirstChunk    = flag.Bool("drop-first-chunk", false, "forego persisting of first received (and typically incomplete) chunk")
-	ingestFromStr     = flag.String("ingest-from", "", "only ingest data with timestamps belonging to the chunk starting after the timestamp specified for an org id; syntax: [ORG_ID]:[TIMESTAMP],[ORG_ID]:[TIMESTAMP],...")
+	ingestFromStr     = flag.String("ingest-from", "", "only ingest data with timestamps belonging to the next chunk from the timestamp specified for an org id; syntax: [ORG_ID]:[TIMESTAMP],[ORG_ID]:[TIMESTAMP],...")
 	chunkMaxStaleStr  = flag.String("chunk-max-stale", "1h", "max age for a chunk before to be considered stale and to be persisted to Cassandra.")
 	metricMaxStaleStr = flag.String("metric-max-stale", "3h", "max age for a metric before to be considered stale and to be purged from memory.")
 	gcIntervalStr     = flag.String("gc-interval", "1h", "Interval to run garbage collection job.")
--- a/cmd/metrictank/metrictank.go
+++ b/cmd/metrictank/metrictank.go
@@ -60,7 +60,7 @@ var (
 
 	// Data:
 	dropFirstChunk    = flag.Bool("drop-first-chunk", false, "forego persisting of first received (and typically incomplete) chunk")
-	ingestFromStr     = flag.String("ingest-from", "", "only ingest data with timestamps belonging to the next chunk from the timestamp specified for an org id; syntax: [ORG_ID]:[TIMESTAMP],[ORG_ID]:[TIMESTAMP],...")
+	ingestFromStr     = flag.String("ingest-from", "", "only ingest data for chunks that have a t0 equal or higher to the given timestamp. Specified per org. syntax: orgID:timestamp[,...]")
 	chunkMaxStaleStr  = flag.String("chunk-max-stale", "1h", "max age for a chunk before to be considered stale and to be persisted to Cassandra.")
 	metricMaxStaleStr = flag.String("metric-max-stale", "3h", "max age for a metric before to be considered stale and to be purged from memory.")
 	gcIntervalStr     = flag.String("gc-interval", "1h", "Interval to run garbage collection job.")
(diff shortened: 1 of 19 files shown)


woodsaj:
why limit this to only 1 orgId?  Whould it not be better to just have AggMetrics use a map of orgId's to ingestAfter timstamp. eg  `ingestAfter map[int32]int32`

Users can then pass a list of org:ts pairs via the config.  eg
`metrictank --ingest-after=1:1562339919,1234:1562339919,45:1562339919`

fkaleo:
> why limit this to only 1 orgId? Whould it not be better to just have AggMetrics use a map of orgId's to ingestAfter timstamp. eg `ingestAfter map[int32]int32`
> 
> Users can then pass a list of org:ts pairs via the config. eg
> `metrictank --ingest-after=1:1562339919,1234:1562339919,45:1562339919`

The only reason was the fact that we did not need it and kept the code a little simpler. We can always add it later.

fkaleo:
> > why limit this to only 1 orgId? Whould it not be better to just have AggMetrics use a map of orgId's to ingestAfter timstamp. eg `ingestAfter map[int32]int32`
> > Users can then pass a list of org:ts pairs via the config. eg
> > `metrictank --ingest-after=1:1562339919,1234:1562339919,45:1562339919`
> 
> The only reason was the fact that we did not need it and kept the code a little simpler. We can always add it later.

Using a map is probably less code though; let me try that.

mdata/aggmetric.go
```diff
@@ -422,6 +430,12 @@ func (a *AggMetric) persist(pos int) {
 
 // don't ever call with a ts of 0, cause we use 0 to mean not initialized!
 func (a *AggMetric) Add(ts uint32, val float64) {
+	if ts < a.ingestAfterNextChunkT0 {
+		// TODO: add metric to keep track of the # of points discarded
+		log.Debugf("AM: discarding metric <%d,%f>: does not belong to a chunk starting after ingest-after. First chunk considered starts at %d", ts, val, a.ingestAfterNextChunkT0)
```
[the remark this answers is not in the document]
fkaleo: I'm curious now: `if logger.IsLevelEnabled(WarnLevel)` is called by `log.Debugf` albeit one function call deeper and with more argument passing: did we notice significant perf improvements with it?
In any case, I added the `if` in the latest commit.

fkaleo:
> I think this might need to be rebased on master

Done

mdata/aggmetric.go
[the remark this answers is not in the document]
shanson7: Yes. You can see my comment in https://github.com/grafana/metrictank/pull/1269

tl;dr The go compiler escape analysis reveals that these lines cause the arguments to escape to the heap, so they result in allocations when called.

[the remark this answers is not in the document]
fkaleo: awesome Sean!

cmd/metrictank/metrictank.go
```diff
@@ -59,6 +60,7 @@ var (
 
 	// Data:
 	dropFirstChunk    = flag.Bool("drop-first-chunk", false, "forego persisting of first received (and typically incomplete) chunk")
+	ingestAfterStr    = flag.String("ingest-after", "", "discards data from an org id received before the next chunk boundary after the specified unix timestamp (in seconds since 1 Jan 1970); syntax: org-id:timestamp")
```
Dieterbe: 1) this is not true in case the specified timestamp is a chunk boundary (than it is not the next chunk)
2) the description of what a unix timestamp is incomplete (UTC timezone), but also this is not the place to explain a unix timestamp
3) would it make sense to have multiple rules for multiple orgs? I thought it did
4) the message is clunky. can we simplify it?

fkaleo: >     1. this is not true in case the specified timestamp is a chunk boundary (than it is not the next chunk)

I thought it was the case given how `ingestAfterNextChunkT0` is computed in `aggmetric.go`:

`ingestAfterNextChunkT0 = uint32(ingestAfter-(ingestAfter%int64(ret.ChunkSpan))) + ret.ChunkSpan`

> 
>     2. the description of what a unix timestamp is incomplete (UTC timezone), but also this is not the place to explain a unix timestamp

Removed.

> 
>     3. would it make sense to have multiple rules for multiple orgs (eg as csv) ? I thought it did

Implemented (546e9aa).

> 
>     4. the message is clunky. can we simplify it?

I took a stab at simplifying it, let me know what you think (689b147).

Dieterbe: > I thought it was the case given how ingestAfterNextChunkT0 is computed in aggmetric.go:

you are correct!

However, I imagine the typical use for a flag like this will be to pass a chunk boundary timestamp, so something like:
```
~ ❯❯❯ TZ=UTC date -d '06:00'
Wed 07 Aug 2019 06:00:00 AM UTC
~ ❯❯❯ TZ=UTC date -d '06:00' +%s
1565157600
~ ❯❯❯ TZ=UTC date -d 'tomorrow 00:00'
Thu 08 Aug 2019 12:00:00 AM UTC
~ ❯❯❯ TZ=UTC date -d 'tomorrow 00:00' +%s
1565222400
```

I don't imagine people would pass a value such as 
```
~ ❯❯❯ TZ=UTC date -d '23:59:59' +%s
1565222399
```

Thus, assuming this hunch is correct and people would pass a clean chunk boundary, i would argue it's confusing if ingestion is only for data that is 1 chunk "after" the timestamp i specified.
Thus perhaps instead of the formula you use, we should use `mdata.AggBoundary()` and change the name "ingest-after" to "ingest-from", describing that is is to ingest data "from this timestamp and onwards" rathar than strictly "after".

WDYT?

Dieterbe: > I took a stab at simplifying it, let me know what you think (689b147).

 :+1:

mdata/aggmetric.go
```diff
@@ -31,39 +31,47 @@ type AggMetric struct {
...
-	firstTs         uint32 // timestamp of first point seen
+	key                    schema.AMKey
+	rob                    *ReorderBuffer
+	currentChunkPos        int    // Chunks[CurrentChunkPos] is active. Others are finished. Only valid when len(chunks) > 0, e.g. when data has been written (excl ROB data)
+	numChunks              uint32 // max size of the circular buffer
+	chunkSpan              uint32 // span of individual chunks in seconds
+	chunks                 []*chunk.Chunk
+	aggregators            []*Aggregator
+	dropFirstChunk         bool
+	ingestAfterNextChunkT0 uint32
```
Dieterbe: another benefit of my suggestion (see other comment) is if the user input is `ingestFrom`, then this var can be `ingestFromT0` and all the alignment here stays the same ;)

util/flag_parsers_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"missing timestamp", "0", 0, 0, true},
+		{"text no number 1", "text", 0, 0, true},
+		{"text no number 2", "1:text", 0, 0, true},
+		{"orgid overflow", "10000000000:0", 0, 0, true},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			defer func() {
+				r := recover()
```
Dieterbe: have a look at : https://golang.org/src/regexp/regexp.go?s=9796:9832#L259
it seems a go best practice to have Must* style functions call regular functions that return an error.
that way in unit tests you can call the non-Must (non-panicing) version and just check for errors, rather than needing recover logic. this seems cleaner and simpler.

fkaleo: > have a look at : https://golang.org/src/regexp/regexp.go?s=9796:9832#L259
> it seems a go best practice to have Must* style functions call regular functions that return an error.
> that way in unit tests you can call the non-Must (non-panicing) version and just check for errors, rather than needing recover logic. this seems cleaner and simpler.

Done in a1c8bf1

mdata/aggmetric.go
fkaleo: > another benefit of my suggestion (see other comment) is if the user input is `ingestFrom`, then this var can be `ingestFromT0` and all the alignment here stays the same ;)

Implemented with 3bb1d5f

cmd/metrictank/metrictank.go
fkaleo: > > I thought it was the case given how ingestAfterNextChunkT0 is computed in aggmetric.go:
> 
> you are correct!
> 
> However, I imagine the typical use for a flag like this will be to pass a chunk boundary timestamp (or, a common enough use), so something like:
> 
> ```
> ~ ❯❯❯ TZ=UTC date -d '06:00'
> Wed 07 Aug 2019 06:00:00 AM UTC
> ~ ❯❯❯ TZ=UTC date -d '06:00' +%s
> 1565157600
> ~ ❯❯❯ TZ=UTC date -d 'tomorrow 00:00'
> Thu 08 Aug 2019 12:00:00 AM UTC
> ~ ❯❯❯ TZ=UTC date -d 'tomorrow 00:00' +%s
> 1565222400
> ```
> 
> I don't imagine people would often pass a value such as
> 
> ```
> ~ ❯❯❯ TZ=UTC date -d '23:59:59' +%s
> 1565222399
> ```
> 
> Thus, assuming this hunch is correct and people would commonly pass a clean chunk boundary, i would argue it's confusing if ingestion is only for data that is 1 chunk "after" the timestamp i specified.
> Thus perhaps instead of the formula you use, we should use [`mdata.AggBoundary()`](https://godoc.org/github.com/grafana/metrictank/mdata#AggBoundary) and change the name "ingest-after" to "ingest-from", describing that is is to ingest data "from this timestamp and onwards" rathar than strictly "after".
> 
> WDYT?


Implemented with commits 67e3d6a and 3a498f4.

mdata/aggmetric.go
```diff
@@ -58,8 +58,8 @@ func NewAggMetric(store Store, cachePusher cache.CachePusher, key schema.AMKey,
 
 	ingestFromT0 := uint32(0)
 	if ingestFrom > 0 {
-		// compute start of next chunk after ingestFrom
-		ingestFromT0 = uint32(ingestFrom-(ingestFrom%int64(ret.ChunkSpan))) + ret.ChunkSpan
+		// compute next chunk boundary starting from 'ingestFrom'
```
Dieterbe: the word "next" seems confusing here.

fkaleo: > the word "next" seems confusing here.

Yeah, I don't know how to phrase it much better :/ And `AggBoundary` uses the word next too :/
`// AggBoundary returns ts if it is a boundary, or the next boundary otherwise.`

Should I remove the comment altogether?

Dieterbe: what i meant is that "next" does not apply in all cases.  AggBoundary comment has the appropriate nuance. anyway i'll give it a shot, no worries.

fkaleo: > what i meant is that "next" does not apply in all cases. AggBoundary comment has the appropriate nuance. anyway i'll give it a shot, no worries.

you're right

mdata/aggregator_test.go
```diff
@@ -91,7 +91,35 @@ func TestAggregator(t *testing.T) {
...
+
+	// chunkspan is 120, ingestFrom = 120 means points before chunk starting at 120 are discarded
+	agg = NewAggregator(mockstore, &cache.MockCache{}, test.GetAMKey(1), ret, aggs, false, 120)
+	agg.Add(100, 123.4)
+	agg.Add(110, 5)
+	agg.Add(130, 130)
+	expected = []schema.Point{
+		{Val: 5, Ts: 120},
+	}
+	compare("simple-min-ingest-from-on-chunk-boundary", agg.minMetric, expected)
```
Dieterbe: too tired now, but tomorrow i'ld like to look into this one last thing, about the timestamps going 130->120 and how that may or should not affect things.

Dieterbe:
LGTM. just need to look at one last thing tomorrow.

mdata/aggregator_test.go
fkaleo: > LGTM. just need to look at one last thing tomorrow.

Thank you! Your changes look great.

mdata/aggmetric_test.go
```diff
@@ -302,6 +302,34 @@ func TestAggMetricDropFirstChunk(t *testing.T) {
...
+	m.Add(22, 22)
+	m.Add(30, 30)
+	m.Add(31, 31)
+	m.Add(32, 32)
+	m.Add(40, 40)
+	itgens, err := mockstore.Search(test.NewContext(), test.GetAMKey(42), 0, 0, 1000)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(itgens) != 1 || itgens[0].T0 != 30 {
```
Dieterbe: what about point with ts=40 ? wouldn't that be in a 2nd chunk?

mdata/aggregator_test.go
```diff
@@ -91,7 +91,35 @@ func TestAggregator(t *testing.T) {
...
+	expected = []schema.Point{}
+	compare("simple-min-ingest-from-all-before-next-chunk", agg.minMetric, expected)
+
+	// chunkspan is 120, ingestFrom = 115 means points before chunk starting at 120 are discarded
+	agg = NewAggregator(mockstore, &cache.MockCache{}, test.GetAMKey(1), ret, aggs, false, 115)
+	agg.Add(100, 123.4)
+	agg.Add(110, 5)
+	agg.Add(130, 130)
+	expected = []schema.Point{
+		{Val: 5, Ts: 120},
```
Dieterbe: where did point ts=130 go?

fkaleo: Comments in 3b4625e should answer that.

mdata/aggmetric_test.go
fkaleo: Comments in 3b4625e should answer that.

fkaleo:
> LGTM
> Please have a close look at the last 2 commits I added.
> If they make sense, merge this.
> If not, let's talk.

They make sense. I'm adding test cases to cover them.

Dieterbe:
I already covered that in the test case i added.
if we need any more tests, it should be one that uses an actual AggMetric with aggregators on it, and confirms the right points make it in to the aggregator (points in the last aggspan before the first t0)

fkaleo:
> I already covered that in the test case i added.
> if we need any more tests, it should be one that uses an actual AggMetric with aggregators on it, and confirms the right points make it in to the aggregator (points in the last aggspan before the first t0)

The test case added does not fail when removing the code changes.
And yep, I'm adding a test to an AggMetric and discovering that in fact more code changes need to happen to make it pass, I think.

fkaleo:
Please take a look at the 2 additional tests and the change in code and let me know what you think.

mdata/aggmetric_test.go
```diff
@@ -302,6 +308,156 @@ func TestAggMetricDropFirstChunk(t *testing.T) {
...
+		t.Fatalf("expected 1 itgens for chunk 30 since the chunks before 25 should be dropped. Got %v", itgens)
+	}
+}
+
+func itersToPoints(iters []tsz.Iter) []schema.Point {
+	points := make([]schema.Point, 0)
```
Dieterbe: https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful

fkaleo: Replaced with nil in commit ad7d980

PR #1387, opened 2019-07-11
fix slow ingest problems

robert-milan:
This PR adds code to the vendored version of raintank/schema which has
already been merged into raintank/schema from this PR:
https://github.com/raintank/schema/pull/33. We can't yet merge master
from raintank/schema due to other required changes. The main point of
the PR is to add and use the Clone() method in raintank/schema/metric.go
in order to assign the nameWithTags values to a cloned MetricDefinition.

We also change the CloneArchive() method in idx/memory.go to use the
new Clone() method. This should reduce allocations caused by CloneArchive()
calling MetricDefinition.NameWithTags() which was causing problems,
especially during startup.

See also: https://github.com/raintank/schema/pull/33

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -705,14 +705,14 @@
   revision = "05a94bb32ad1f23f4b01edb2edd06862d4a484d2"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


robert-milan:
Tests are passing, it is failing QA due to a difference in vendored code being changed here locally.

PR #1418, opened 2019-08-06
[WIP] Single name interning

robert-milan:


robert-milan:
to revisit / re-use later

PR #1434, opened 2019-08-15
move schema into metrictank

robert-milan:
This moves the vendored code from `raintank/schema` into `Metrictank`. Other projects will need to be updated after this is merged.

After this is merged the documentation at godoc.org will be udpated and reflect what is currently in the documentation found at `docs/inputs.md`


woodsaj:
Looks like `go vet` or `gofmt` changes have been applied to everything in vendor/,  that should be reverted.

Otherwise looks good to me.

robert-milan:
> Looks like `go vet` or `gofmt` changes have been applied to everything in vendor/, that should be reverted.
> 
> Otherwise looks good to me.

Yes they were, I am just undoing that and updating `go generate` right now

PR #1445, opened 2019-08-27
add missing test file to schema/msg

robert-milan:
Forgot to add this test file when moving schema from raintank into here.

Closes #1443


robert-milan:
Disregard

PR #1543, opened 2019-11-20
Added comment as to why we upgraded to a certain version of gocql

fkaleo:


fkaleo:
gocql upgrade was reverted

PR #1608, opened 2020-01-15
mt-gateway: support ingesting data into kafka

fitzoh:
Initial portion of this PR was porting the `github.com/raintank/tsdb-gw/publish` package to this repository for use in a new HTTP proxy (see https://github.com/grafana/metrictank-ops/issues/542 if you have access).

The portion after [this comment](https://github.com/grafana/metrictank/pull/1608#issuecomment-575830102) (specifically after the force push) pulls in the actual API endpoint and wires it up to Kafka.

A couple design decisions:
* This is intended to be a simplified API, so rate limiting functionality has been removed.
* As there is no concept of authentication/authorization within metrictank, the `isAdmin=true` path is always taken.

Once again, the latter portion of this PR is intended to be reviewed commit by commit.

Open questions:
* Do we need to consider any licensing/attribution issues?
    * probably fine, could use a +1 from @woodsaj ✔️
* Is there any dead code we can get rid of post-import?
   * removed ✔️
* Should we maintain the instrumentation present in the original code?
    * leave it ✔️
* Testing? Do we add this to a docker config? Which one?
    * toss it in `docker-dev-custom-cfg-kafka` alongside `tsdb-gw` for now


publish/publish.go
```diff
@@ -0,0 +1,75 @@
...
+type Publisher interface {
+	Publish(metrics []*schema.MetricData) error
+	Type() string
+}
+
+var (
+	publisher Publisher
+
+	// Persister allows pushing metrics to the Persistor Service
+	Persistor *metrics_client.Client
```
fitzoh: It doesn't look like this `Persistor` is actually used (either in this PR or anywhere in the tsdb-gw project), which I believe means we can drop it as well as the `metrics_client` package which was pulled in specifically for this.

publish/kafka/schemas.go
```diff
@@ -0,0 +1,13 @@
+package kafka
+
+import (
+	"github.com/grafana/metrictank/conf"
+)
+
+func getSchemas(file string) (*conf.Schemas, error) {
```
fitzoh: This function and file can probably go away

publish/publish.go
replay: hmm i have a feeling that this was added because historically we used to use tsdb-gw also to ingest data into cortex, but now we don't do that anymore afaik

replay: so `metrics_client` is only necessary for the `persistor`? in that case i think we can remove all of it probably

publish/kafka/schemas.go
replay: why can this go away? don't we need this when initializing the `kafka.mtPublisher`, to set its `schemas` property (https://github.com/grafana/metrictank/pull/1608/files#diff-c1de7345bb1a47b5638ddad84e8ae8eeR200)?

fitzoh: Screen Shot 2020-01-16 at 1 12 21 PM

publish/publish.go
fitzoh: 3af36e930d997bc3facbc9d115d4ba5401c00ed5

publish/kafka/schemas.go
fitzoh: removed in 34f4d3884938feb1e9b80211561a684afa5f1379

replay:
To be sure I understand this diff right, this doesn't change the Sarama version, it only adds another package right? I'm asking because when changing Sarama versions it's usually a good idea to do some benchmarks to verify that there are no regressions.

replay:
And currently none of this code which gets added is actually being used, right? That's all going to be used by your new http gateway?

replay:
By merging the dependencies of the mt-gateway into the MT repo separately before they actually get used, isn't there a risk that we accidentally copy more stuff from tsdb-gw than what's actually going to be used and this will then be left there as dead code, or are you sure that all of this is needed?

robert-milan:
> To be sure I understand this diff right, this doesn't change the Sarama version, it only adds another package right? I'm asking because when changing Sarama versions it's usually a good idea to do some benchmarks to verify that there are no regressions.

Yes that is correct. I thought the same thing at first glance, but it only adds `mocks`.

I understand that breaking it up like this makes for smaller and easier to read PRs, but in this case it does seem a little bit awkward. I think this should be part of the PR that uses this code in the HTTP proxy. Since #1602 is already merged, can we just rename this PR and add the HTTP proxy changes here as well? It is easy enough to review commit by commit.

> Do we need to consider any licensing/attribution issues?

We probably should take a look at everything

> Are we able to omit any portion of this package?

That would be easier to determine once we see some of the implementation

> Should we maintain the instrumentation present in the original?

I don't think we need to maintain the instrumentation unless there is a good reason for it.

fitzoh:
>>To be sure I understand this diff right, this doesn't change the Sarama version, it only adds another package right? I'm asking because when changing Sarama versions it's usually a good idea to do some benchmarks to verify that there are no regressions.

>Yes that is correct. I thought the same thing at first glance, but it only adds mocks.

Correct, @robert-milan  


>And currently none of this code which gets added is actually being used, right? That's all going to be used by your new http gateway?

>I understand that breaking it up like this makes for smaller and easier to read PRs, but in this case it does seem a little bit awkward. I think this should be part of the PR that uses this code in the HTTP proxy. Since #1602 is already merged, can we just rename this PR and add the HTTP proxy changes here as well? It is easy enough to review commit by commit.


Yeah, good call.  This was mostly an attempt to unblock myself while #1602 was waiting on reviews... now that it's merged, I'll rebase and work on wiring things up wire things up

cmd/mt-gateway/ingest/metrics.go
```diff
@@ -116,109 +117,124 @@ func prepareIngest(ctx *models.Context, in []*schema.MetricData, toPublish []*sc
...
-		case <-ctx.Req.Context().Done():
-			ctx.Error(499, "request canceled")
+		case <-r.Context().Done():
+			w.WriteHeader(499)
+			fmt.Fprintf(w, "request canceled")
 		default:
 			log.Errorf("unable to read request body. %s", err)
-			ctx.JSON(500, err)
+			w.WriteHeader(500)
+			fmt.Fprintf(w, "unable to read request body. %s", err)
```
replay: there's a repeating pattern of writing a status code header, then writing a message on the following line, and in some cases also log a line with the same msg. 
would it make sense to put these into a small helper like `writeStatus(w, 499, "request cancelled", true)` (the last param could determine whether to log).

cmd/mt-gateway/api.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/grafana/metrictank/cmd/mt-gateway/ingest"
```
replay: that's just a convention: we usually sort the imports so that there are first all the standard library imports, then an empty line, then all the imports of other pkgs. (f.e. https://github.com/grafana/metrictank/blob/master/idx/memory/memory.go#L3)

cmd/mt-gateway/api.go
```diff
@@ -21,13 +24,28 @@ type Api struct {
...
 	api.graphiteHandler = withMiddleware("graphite", httputil.NewSingleHostReverseProxy(urls.graphite))
 	api.metrictankHandler = withMiddleware("metrictank", httputil.NewSingleHostReverseProxy(urls.metrictank))
 	api.bulkImportHandler = withMiddleware("bulk-importer", bulkImportHandler(urls))
 	return api
 }
 
+func ingestHandler(urls Urls) http.Handler {
+	publisher := kafka.New(strings.Split(urls.kafkaBrokers, ","), true)
+	if publisher == nil {
+		log.Info("metrics ingestion not enabled")
```
replay: it might be helpful to new users if the message mentions that metrics ingestion is not enabled because no kafka brokers were configured.

cmd/mt-gateway/api.go
```diff
@@ -21,13 +24,28 @@ type Api struct {
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			_, _ = fmt.Fprintln(w, "no url configured for bulk importer service")
```
replay: this message seems wrong

replay (changes requested):
Looks good, just a few minor comments

cmd/mt-gateway/api.go
fitzoh: 87cb16c62265e5574e616489cd4489007c8f53e6

cmd/mt-gateway/api.go
fitzoh: 87cb16c62265e5574e616489cd4489007c8f53e6

cmd/mt-gateway/api.go
fitzoh: 186fab2a2c2d42fa4de174c437a2f11361fb69c2

cmd/mt-gateway/ingest/metrics.go
fitzoh: Yeah, that had crossed my mind.
Skipped the  boolean and just always log @ error (successful ingestions take a different code path)

1abc592486eb7c50c4d2b25b84f73d6c037704dc

cmd/mt-gateway/ingest/metrics.go
```diff
@@ -227,3 +207,10 @@ func metricsBinary(w http.ResponseWriter, r *http.Request, compressed bool) {
 	w.WriteHeader(200)
 	json.NewEncoder(w).Encode(resp)
 }
+
+func writeErrorResponse(w http.ResponseWriter, status int, msg string, fmtArgs ...interface{}) {
+	w.WriteHeader(status)
+	formatted := fmt.Sprint(msg, fmtArgs)
+	log.Error(formatted)
+	fmt.Fprintf(w, formatted)
```
replay: this could be `fmt.Fprint()` now, instead of `Fprintf`

fitzoh: woops, 3237276fcfbf60999ddfa8328c93db1abb686e9c

replay (approved):
LGTM

PR #1638, opened 2020-01-27
Revert to smaller CI instance size

fitzoh:
There was concern about angering the CircleCI billing gods if we stayed at the larger instance size, and we're not entirely sure it's necessary.

Let's see if this PR passes it's checks.


fitzoh:
Green on the first try, giving it one more rerun

PR #1642, opened 2020-01-28
Add ini file for mt-gateway

fitzoh:
Add an ini file similar to the one used for metrictank and tsdb-gw to the new mt-gateway.

Also helps to provide a bit of additional documentation around configuration options.

### DIFF
--- a/cmd/mt-gateway/main.go
+++ b/cmd/mt-gateway/main.go
@@ -3,17 +3,21 @@ package main
 import (
 	"flag"
 	"fmt"
-	"github.com/grafana/metrictank/logger"
-	log "github.com/sirupsen/logrus"
 	"net/http"
 	"net/url"
+	"os"
 	"runtime"
+
+	"github.com/grafana/globalconf"
+	"github.com/grafana/metrictank/logger"
+	log "github.com/sirupsen/logrus"
 )
 
 var (
 	version       = "(none)"
 	showVersion   = flag.Bool("version", false, "print version string")
 	addr          = flag.String("addr", ":6059", "http service address")
+	confFile      = flag.String("config", "/etc/metrictank/mt-gateway.ini", "configuration file path")
 	metrictankUrl = flag.String("metrictank-url", "http://localhost:6060", "metrictank address")
 	graphiteURL   = flag.String("graphite-url", "http://localhost:8080", "graphite-api address")
 	importerURL   = flag.String("importer-url", "", "mt-whisper-importer-writer address")
@@ -47,14 +51,27 @@ func main() {
 		fmt.Println("Flags:")
 		flag.PrintDefaults()
 	}
-	flag.Parse()
+
+	// Only try and parse the conf file if it exists
+	path := ""
+	if _, err := os.Stat(*confFile); err == nil {
+		path = *confFile
+	}
+	conf, err := globalconf.NewWithOptions(&globalconf.Options{
+		Filename:  path,
+		EnvPrefix: "MT_",
+	})
+	if err != nil {
+		log.Fatalf("error with configuration file: %s", err)
+		os.Exit(1)
+	}
+	conf.ParseAll()
 
 	if *showVersion {
 		fmt.Printf("mt-gateway (version: %s - runtime: %s)\n", version, runtime.Version())
 		return
 	}
 
-	var err error
 	urls := Urls{}
 	urls.kafkaBrokers = *brokers
 
--- a/cmd/mt-gateway/main.go
+++ b/cmd/mt-gateway/main.go
@@ -63,7 +63,6 @@ func main() {
 	})
 	if err != nil {
 		log.Fatalf("error with configuration file: %s", err)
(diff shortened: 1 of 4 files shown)


scripts/config/mt-gateway.ini
```diff
@@ -0,0 +1,63 @@
...
+only-org-id = 0
+
+#discard data points starting with one of the given prefixes separated by | (may be given multiple times, once per topic, as a comma-separated list)
+#discard-prefixes =
+
+#enable optimized MetricPoint payload
+v2 = true
+
+#interval after which we always resend a full MetricData
+v2-clear-interval duration = 1h0m0s
```
replay: is this missing a `-`

cmd/mt-gateway/main.go
```diff
@@ -47,14 +51,27 @@ func main() {
...
-	flag.Parse()
+
+	// Only try and parse the conf file if it exists
+	path := ""
+	if _, err := os.Stat(*confFile); err == nil {
+		path = *confFile
+	}
+	conf, err := globalconf.NewWithOptions(&globalconf.Options{
+		Filename:  path,
+		EnvPrefix: "MT_",
```
replay: would it make sense to use a different string than `MT_`? IIRC in tsdb-gw we're using `GW_`, maybe that would make more sense since this is also a gateway.

cmd/mt-gateway/main.go
```diff
@@ -47,14 +51,27 @@ func main() {
...
+	})
+	if err != nil {
+		log.Fatalf("error with configuration file: %s", err)
+		os.Exit(1)
```
replay: is the `os.Exit()` reached after `log.Fatalf()`? I thought `log.Fatalf()` should already exit the program

cmd/mt-gateway/main.go
fitzoh: I was going back and forth of that...

I think it's pretty likely that we'll be sharing configuration (we're using the same config dir and likely schema files), so we'll probably want to share env vars as well.

I could be convinced though

cmd/mt-gateway/main.go
fitzoh: You are correct: 2a9acc35ef9331d5cfe286152a6e83b2dad9a108

cmd/mt-gateway/main.go
replay: fair enough. at least that would then be consistent with what the documentation says https://github.com/grafana/metrictank/blob/master/docs/config.md#config

scripts/config/mt-gateway.ini
fitzoh: I'm actually not sure what happened there: 6bb9d11b1ba7cf8bdce09ee4fd5a61d41786f2b7

cmd/mt-gateway/main.go
```diff
@@ -47,14 +51,26 @@ func main() {
...
+	path := ""
+	if _, err := os.Stat(*confFile); err == nil {
+		path = *confFile
+	}
+	conf, err := globalconf.NewWithOptions(&globalconf.Options{
+		Filename:  path,
+		EnvPrefix: "MT_",
+	})
+	if err != nil {
+		log.Fatalf("error with configuration file: %s", err)
```
replay: the message is saying there's an error with the configuration file, but couldn't it also be an error with the env variable values? 

```suggestion
		log.Fatalf("error with configuration: %s", err)
```

scripts/config/mt-gateway.ini
```diff
@@ -0,0 +1,63 @@
...
+### The properties below are for configuring ingestion to kafka via the `/metrics` endpoint
+###
+
+#enable metric publishing
+metrics-publish = false
+
+#path to carbon storage-schemas.conf
+schemas-file = /etc/metrictank/storage-schemas.conf
+
+#kafka tcp address(es) for metrics, in csv host[:port] format
```
replay: wouldn't `csv` mean that it's ok to have spaces after the `,`? Do we allow spaces after the `,`?

cmd/mt-gateway/main.go
fitzoh: Impl looks like file loading is the only thing that can generate an error
Screen Shot 2020-01-28 at 12 10 22 PM

scripts/config/mt-gateway.ini
fitzoh: How do you feel about dc6254a242c77e3545ba92b3d931ce5b7f4e79af ?

Worst case we should fail fast and they can figure it out from the error message

replay: I'd just say it the same was as in the metrictank config, for consistency and clarity: https://github.com/grafana/metrictank/blob/master/metrictank-sample.ini#L271

fitzoh: 0c55a44ee243c3794318a43f597ef6abac45fd3d

replay (approved):
LGTM

PR #1654, opened 2020-01-30
Delegate `Delete` calls to `DeleteTagged` if query is `seriesByTag`

fitzoh:
Update `UnpartitionedMemoryIdx.Delete` to delegate to `UnpartitionedM…emoryIdx.DeleteTagged` if the passed in query is a `seriesByTag` query.

The comments in #1023 state that `/metrics/delete` should function similarly to `tags/delSeries` if a `seriesByTag` query is passed in to it.

This PR accomplishes that by adding a hook in `UnpartitionedMemoryIdx.Delete`, which all other `MetricIndex.Delete` calls appear to pass through eventually.

Does not currently compile, depends on changes in #1652.

Tests not yet implemented (related test fixes waiting for review in #1653), want to make sure this approach makes sense first.

### DIFF
--- a/idx/memory/memory.go
+++ b/idx/memory/memory.go
@@ -1829,18 +1829,34 @@ func (m *UnpartitionedMemoryIdx) Delete(orgId uint32, pattern string) ([]idx.Arc
 			}
 		}
 	}()
-	tree, ok := m.tree[orgId]
-	if !ok {
-		return nil, nil
-	}
-	found, err := find(tree, pattern)
-	if err != nil {
-		return nil, err
-	}
 
-	for _, f := range found {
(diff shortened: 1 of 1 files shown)


fitzoh:
Closing based on some out of band feedback, new version coming soon:

>in general i think it's a good idea to keep as much logic outside the index as possible, especially parsing and request preparation etc. the index is already complicated enough

>So are you thinking the delegation to `DeleteTagged` should happen further up the stack (pre-index)?

>yeah, i'd make the decision which method of deletion to use before the index, and then just call the correct index method

>so `Server.metricsDeleteLocal` and `Server.IndexDelete`

> yeah

PR #1686, opened 2020-02-25
Expose method to force a trace to be sampled

shanson7:
We receive several hundred requests per second on our cluster and because we have a large number of read instance, each trace is several hundred (or sometimes thousand!) spans large. This means we use the probabilistic sampler to sample only a small number of requests. This works well, but it means that when we want to sample particular requests (for example, when debugging or doing some benchmarking) it is cumbersome to get a request that is sampled.

### DIFF
--- a/api/middleware/tracer.go
+++ b/api/middleware/tracer.go
@@ -7,6 +7,7 @@ import (
 	"github.com/grafana/metrictank/tracing"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	log "github.com/sirupsen/logrus"
 	jaeger "github.com/uber/jaeger-client-go"
 	"gopkg.in/macaron.v1"
 )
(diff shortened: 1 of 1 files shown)


woodsaj:
I dont think that this is a good idea.

Metrictank wont always be the first process which receives a request and creates the root span (it is not in our deployments).  This will result in the root span being dropped.

You can already force a trace by including the `jaeger-debug-id` header in a request.  the value of the header is used as the trace-id.

shanson7:
> Metrictank wont always be the first process which receives a request

That's fine. Whatever gets the initial request, if trace enabled, can send the trace details along instead.

> This will result in the root span being dropped.

Not sure I understand. I don't think this would drop the root span, the root span may just not have been sampled (in the case that for whatever reason the thing generating the root span sent this flag)

> You can already force a trace by including the jaeger-debug-id header in a request. the value of the header is used as the trace-id.

IMO this is much stranger and more difficult than a simple flag. Many tools that allow custom URLs don't allow custom headers or it's more difficult to use them. This seems like it would also have all the negatives you just mentioned, while being harder to use.

shanson7:
Playing around with the header method, it looks like the trace-id is still generated by the system (which is good, since that means we can use a static correlation id). This is probably sufficient for this use case, just excluding a few simpler tools that don't allow specifying headers easily (e.g. the browser).

shanson7:
Oh, I should also mention that this branch fixed the "-local" trace behavior by parsing the form first. Seems like it stopped working at some point.

PR #1691, opened 2020-02-26
Tags support for Carbon output

fkaleo:


cmd/mt-fakemetrics/out/carbon/carbon.go
```diff
@@ -62,7 +62,17 @@ func (n *Carbon) Flush(metrics []*schema.MetricData) error {
 	preFlush := time.Now()
 	buf := bytes.NewBufferString("")
 	for _, m := range metrics {
-		buf.WriteString(fmt.Sprintf("%s %f %d\n", m.Name, m.Value, m.Time))
+		tagsString := ""
+		for i, tag := range m.Tags {
+			if i == 0 {
+				tagsString += ";"
+			} else {
+				tagsString += ","
```
fitzoh: pretty sure it should always be `;`
Example from https://graphite.readthedocs.io/en/latest/tags.html:
`my.series;tag1=value1;tag2=value2`

fkaleo: sorry yes; I don't know why I thought `,` was needed

fkaleo: fixed in 8db6fb3

PR #1698, opened 2020-02-28
yet another WIP refactor for stats tagging that stores the name/tags within the metric itself

fitzoh:
Alternative to https://github.com/grafana/metrictank/pull/1696 that stores the name/tags within the metric, allowing it to be written without having to pass it's own name in.

Also includes a separate tag friendly constructor.

This version doesn't compile, just demonstrating what metrics/API look like to validate before applying more broadly

### DIFF
--- a/stats/bool.go
+++ b/stats/bool.go
@@ -6,11 +6,22 @@ import (
 )
 
 type Bool struct {
-	val uint32
+	val  uint32
+	name []byte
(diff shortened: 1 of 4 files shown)


fitzoh:
Going with this approach as discussed in #1696 (although probably with a single constructor w/ optional tags), closing until ready for review

PR #1823, opened 2020-05-13
Fix remove empty series

shanson7:
Noticed that removeEmptySeries was still proxying to graphite (due to requiring the xFilesFactor arg). While looking into that, I noticed that the output is different from graphite, so fixed that as well.

### DIFF
--- a/expr/func_removeemptyseries.go
+++ b/expr/func_removeemptyseries.go
@@ -1,7 +1,6 @@
 package expr
 
 import (
-	"fmt"
 	"math"
 
 	"github.com/grafana/metrictank/api/models"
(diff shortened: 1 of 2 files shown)


shanson7:
cc @vaguiar

expr/func_removeemptyseries.go
```diff
@@ -42,9 +41,6 @@ func (s *FuncRemoveEmptySeries) Exec(dataMap DataMap) ([]models.Series, error) {
 
 	var output []models.Series
 	for _, serie := range series {
-		serie.Target = fmt.Sprintf("removeEmptySeries(%s, %g)", serie.Target, s.xFilesFactor)
```
vaguiar, line 45: With the default being set in the constructor, we might not need the default override in line 39

PR #1830, opened 2020-05-20
[DNM] Native implementation of movingWindow and children

vaguiar:
### Describe your changes
This PR implements the movingWindow and it's children functions like movingAverage etc. natively. Note: There's no support for fetching a time slice based on number of points. 

I've created this PR for early feedback. This can't be merged until https://github.com/grafana/metrictank/pull/1739 goes through

### Testing performed
- A/B testing with the mtcmptest tool. 
- Unit tests
- Integration tests

### Additional context
- I've seen some mismatches for the aggregate value calculations for some datapoints, while others in the series matching perfectly. (possibly due to the differences due to post-marking in graphite vs metrictank?)
- I've deleted the old (unstable) implementation of movingAverage 
- Wrote a common function to do xFilesFactor checks for a series of points which is used here and in removeEmptySeries


shanson7:
This isn't ready yet. It depends on (and erroneously includes commits from) #1739

PR #1837, opened 2020-05-28
Use non-blocking write to flushed

### LINKED ISSUE #1836: Deadlock in memory/write_queue
**Describe the bug**
The scenario is described below:

Several goroutines are in the [Queue](https://github.com/grafana/metrictank/blob/master/idx/memory/write_queue.go#L46) function, all but 1 stuck on the Lock call. The one that gets the Lock also flushes because it hits the max buffer size

The loop thread is waiting on Lock as well in the [timer](https://github.com/grafana/metrictank/blob/master/idx/memory/write_queue.go#L94)

After the flush, `wq.flushed` has an element in it (and is therefore full)

One of the Queue routines acquires the lock and adds an archive so `len(wq.archives) == 1`

The loop thread now acquires the lock, and executes it's flush. But at the end of the flush, it tries to push into `wq.flushed` and is blocked

Simple solution is to use a non-blocking write to `wq.flushed`. 

**Helpful Information**
Metrictank Version: master

shanson7:
Fix #1836 

Simplest change is to use a non-blocking write to the `flushed` chan. Possibly more ideal would be for the `Queue`ing thread to trigger a flush, rather than flush manually, but that would be a larger change that would need more proof of correctness.

### DIFF
--- a/idx/memory/write_queue.go
+++ b/idx/memory/write_queue.go
@@ -79,9 +79,9 @@ func (wq *WriteQueue) flush() {
 	wq.archives = make(map[schema.MKey]*idx.Archive)
 
 	select {
-       case wq.flushed <- struct{}{}:
-       default:
-    }
+	case wq.flushed <- struct{}{}:
+	default:
+	}
 }
 
 func (wq *WriteQueue) loop() {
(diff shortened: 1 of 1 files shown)


shanson7:
Tests are hitting the timeout. Just running the memory tests locally takes over a minute, almost entirely in `memory_find_test.go`.

With additional testing, perhaps we should raise the timeouts or run some tests in parallel. Either way, succeeded in 9 min 51 sec (9 whole seconds to spare!) on the second run.

Issue #1885, opened 2020-08-21
dont hold write lock for entire duration of index delete calls


woodsaj:
When an index Delete is called, the entire Delete operation is performed while a write lock is held.  While the lock is held, no metrics can be ingested and no other index searches can run.

To avoid problems , the Delete function https://github.com/grafana/metrictank/blob/master/idx/memory/memory.go#L1492  should instead
1) use a read lock and perform a find() to get all matching nodes, then release the read lock.
2) for each matching node, acquire a read lock, then traverse the tree to find all matching leaf nodes, then release the read lock.
3) for each matching leaf node, acquire a write lock, then delete the leaf node, which will cascade down the tree and clean up any branches that have no children.

eg something like

```
	var deletedDefs []idx.Archive
	pre := time.Now()
	defer func() {
		if len(deletedDefs) == 0 {
			return
		}
		if m.findCache == nil {
			return
		}
		// asynchronously invalidate any findCache entries
		// that match any of the deleted series.
		if len(deletedDefs) > findCacheInvalidateQueueSize {
			m.findCache.Purge(orgId)
		} else {
			for _, d := range deletedDefs {
				m.findCache.InvalidateFor(orgId, d.NameWithTags())
			}
		}
	}()
	m.RLock()
	tree, ok := m.tree[orgId]
	if !ok {
                m.RUnlock()
		return nil, nil
	}

	found, err := find(tree, pattern)
        m.RUnlock()
	if err != nil {
		return nil, err
	}
        leafNodes := make([]*Node, 0)
	for _, node := range found {
                m.RLock()
                // GetLeaves would be some recursive function like https://github.com/grafana/metrictank/blob/master/idx/memory/memory.go#L1537-L1552
		leafNodes = append(leafNodes, node.GetLeaves()) 
                m.RUnlock()
	}
        for _, node := range leafNodes {
                m.Lock() // should probably use the same lock rate limiting that is used in Prune
                deleted := m.delete(orgId, node, true, false)
		deletedDefs = append(deletedDefs, deleted...)
                m.Unlock()
        }
	statMetricsActive.DecUint32(uint32(len(deletedDefs)))
	statDeleteDuration.Value(time.Since(pre))

	return deletedDefs, nil
}
```

fkaleo:
https://github.com/grafana/metrictank/pull/1897 might fix it, if not @robert-milan has another branch that should and will do a PR for

PR #1897, opened 2020-09-01
do not allow indefinite write lock during deletes

robert-milan:
Changes the behavior of delete operations to prevent them from completely locking up the index for extended periods of time

I initially tried getting all the leaves and deleting one by one, but this proved to be far too slow. I have removed that code but you can see in the commits.

Will this be enough to alleviate the issues of locking the index for too long though?

Fixes: #1885

### DIFF
--- a/idx/memory/memory.go
+++ b/idx/memory/memory.go
@@ -264,6 +264,25 @@ func (n *Node) String() string {
 	return fmt.Sprintf("branch - %s", n.Path)
 }
 
+func (n *Node) GetLeaves(tree *Tree) []*Node {
+	var nodes []*Node
+	if n.HasChildren() {
+		for _, child := range n.Children {
+			node, ok := tree.Items[n.Path+"."+child]
+			if !ok {
+				corruptIndex.Inc()
+				log.Errorf("memory-idx: node %q missing. Index is corrupt.", n.Path+"."+child)
+				continue
+			}
+			nodes = append(nodes, node)
+			if node.HasChildren() {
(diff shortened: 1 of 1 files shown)


idx/memory/memory.go
```diff
@@ -1543,18 +1541,32 @@ func (m *UnpartitionedMemoryIdx) Delete(orgId uint32, pattern string) ([]idx.Arc
...
-		deleted := m.delete(orgId, f, true, true)
+	bc := BlockContext{}
+
+	// limit delete operations holding a write lock to 200ms of every second
+	tl := NewTimeLimiter(time.Second, time.Millisecond*200, time.Now())
+
+	for _, node := range found {
+		tl.Wait()
+		lockStart := time.Now()
+		bc = m.Lock()
```
shanson7, line 1565: Maybe I'm misunderstanding something here, but it seems like the write lock is being acquired and released in each iteration. Perhaps it should only be released when it has used up it's time slice?

Repeatedly acquiring write locks is very expensive in the memory index.

robert-milan: I am still experimenting with deleting by leaf. I don't think this current approach solves the entire problem set. If someone tries to delete 'toplevel.*' when they have a lot of child nodes it will still lock up the index since the current method would recursively call all the way down the tree.

woodsaj: That is correct, deleting `*` would be even worse.

We definitely need the delete call to find all leaf nodes to be deleted while only holding read locks.  Then delete these in batches with write locks.

PR #1911, opened 2020-09-30
change defById to per-org and adds per-org locking to the index

robert-milan:
This changes the nature of defById from a flat map to a per-org map. It also adds per-org locking which may result in improved performance on multitenant instances and add a layer of protection between the orgs.

Overall I am happy-ish with this PR, but I believe that I will take another look at the write_queue and partitioned index. I don't like the extra checks during `add` and `Queue` so I am open to ideas.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ jobs:
       - image: circleci/golang:1.15.2
     steps:
       - checkout
-      - run: go test -v -race $(go list ./... | grep -v github.com/grafana/metrictank/stacktest)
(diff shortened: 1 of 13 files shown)


shanson7:
I would strongly recommend against pushing this into the existing metric index component. The locking here is already very complex and adding more locks is just exacerbating the problem. If you want a per org index, maybe build it on top of `UnpartitionedMemoryIdx ` similar to the partitioned index?

fkaleo:
Let's postpone this work for now.

PR #1948, opened 2020-11-14
Fix set tags for complex targets

### LINKED ISSUE #1917: groupByTags tag parsing adds bad tags
**Describe the bug**

[groupByTags](https://github.com/grafana/metrictank/blob/108ffb9ea16eb92bcfe0bd16002c64fc8b14fe0b/expr/func_groupbytags.go#L85) builds a buffer of `;` delimited tags to form the `key` and then groups all series with the same key together. To get the individual tag values back, it calls [SetTags()](https://github.com/grafana/metrictank/blob/b8bb52605ce405b361c54e2ae5f02c0e00362b6c/api/models/series.go#L122) on the result series. 

Given input from a "complex" pipeline like:

```
groupByTags(
  divideSeriesLists(
    sortByName(
      seriesByTag('namespace=os', 'cluster=clusterName', 'name=cpu.percent.user'),
    false), 
    removeBelowValue(
      sortByName(
        seriesByTag('namespace=os', 'cluster=clusterName', 'name=cpu.percent.idle'),
      false),
    10)
  ),
'sum', 'name')
```

and let's say the input series just has the cluster tag. After going through `SetTags()` we get tags like:
`name=divideSeries(cpu.percent.user.g`
`cluster=clusterName`
`namespace=os, 10))`

groupByTags *should* have just produced the `name` tag but because it calls `SetTags()` and the target after divideSeriesLists had values like `divideSeries(cpu.percent.user;cluster=clusterName;namespace=os,removeBelowValue(cpu.percent.idle;cluster=clusterName;namespace=os, 10))` it just naively splits on `;` and adds some bogus tags.

fkaleo:
When a target is more complex than just containing a raw series, ie. when it contains at least one function, then `SetTags` should consider as tags only the tag expressions (ie. `;tag=value`) that are found after the end of the last function of the target.

Fix #1917

### DIFF
--- a/api/models/series.go
+++ b/api/models/series.go
@@ -120,7 +120,8 @@ func (a SeriesMeta) CopyWithChange(fn func(in SeriesMetaProperties) SeriesMetaPr
 // SetTags fills in the Tags property with tags parsed out of s.Target
 // the "name" tag is always set from s.Target, overriding any attempted override
 func (s *Series) SetTags() {
-	numTags := strings.Count(s.Target, ";")
+	lastClosingParenthesis := strings.LastIndexByte(s.Target, ')')
+	numTags := strings.Count(s.Target[lastClosingParenthesis+1:], ";")
 
 	if s.Tags == nil {
 		// +1 for the name tag
--- a/api/models/series_test.go
+++ b/api/models/series_test.go
@@ -132,6 +132,14 @@ func TestSetTags(t *testing.T) {
 		},
 		{
 			in: Series{
+				Target: "func(a;b=c,d;e=f)",
+			},
+			out: map[string]string{
+				"name": "func(a;b=c,d;e=f)",
+			},
+		},
+		{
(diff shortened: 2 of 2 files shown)


api/models/series.go
```diff
@@ -120,7 +120,8 @@ func (a SeriesMeta) CopyWithChange(fn func(in SeriesMetaProperties) SeriesMetaPr
 // SetTags fills in the Tags property with tags parsed out of s.Target
 // the "name" tag is always set from s.Target, overriding any attempted override
 func (s *Series) SetTags() {
-	numTags := strings.Count(s.Target, ";")
+	lastClosingParenthesis := strings.LastIndexByte(s.Target, ')')
+	numTags := strings.Count(s.Target[lastClosingParenthesis+1:], ";")
```
shanson7, line 124: In the scenario where `numTags >= 1` I don't believe this solution works.

PR #1963, opened 2021-02-23
add Method.String() and NewMethod() methods

Dieterbe:
to clean up code and make conf.Method useful in other projects


fkaleo (approved):
After careful inspection, it looks strictly equivalent. Good to go.

PR #2015, opened 2021-11-30
Move timezone and utils to separate package

djedruszczak:
This PR moves the `api.timeZone` var and certain utilities from the `api` package into a new `tz` package. The `GetFromTo` function is exported for use in a future native implementation of the [`linearRegression` function](https://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.linearRegression).

Some encapsulation of the `api.timeZone` var (now `tz.TimeZone`) is lost so that it can continue to be set from `api.ConfigSetup`, maintaining backwards compatibility. The `api` package is highly dependent on the `expr` package, making the export of `api.getFromTo` from the `api` package for use in native function implementations (the `expr` package) difficult due to import cycles.


Dieterbe (commented):
in master, the timezone is considered an api level configuration, as various of the api methods - before involving expr - call getFromTo, and thus need access to the default timezone configured.
I'm not sure why linearRegression would need to access that variable.  Is it because from/to can be any 'at' style expression, which allows specifying a timezone?  wouldn't that timezone then be scoped to that function call only? it has no bearing on the request-scoped from/to, correct? do you envision needing only read access or also writing to the timezone variable?

also, while i never particularly liked the idea of bringing all api models together in one 'models' package, that is how it is, and moving just a single model ( FromTo ) elsewhere is definitely counter to our current code structure.

robert-milan:
I'll be taking over reviewing this for now

robert-milan (approved):
Yes, we are slightly departing from our traditional layout of Metrictank files / models / separation of code, however I think we should just move forward with this. If it becomes a problem in the future we can always change it.

robert-milan:
Can we get @djedruszczak to agree to the CLA?

djedruszczak:
> Can we get @djedruszczak to agree to the CLA?

Signed!

PR #2016, opened 2021-12-02
Fix benchmarks

djedruszczak:
Most if not all benchmarks were failing because the test series had an interval of 0 -- causing a panic when dividing by 0 when normalizing the series.

This PR modifies the test functions to return the interval as well as the datapoints.


Dieterbe (approved):
nice one!

PR #2038, opened 2022-07-29
Harden index load against failures on startup

shanson7:
After a cassandra crash, invalid index entries caused MT to crash repeatedly. Additionally, there were spurious timeouts and metrictank cannot start if *any* partition failed to load. With even a low chance of partition failure (e.g. 10%) and 16 partitions to load, it's likely Metrictank will fail to start several times.  Add some retries so that it's much more likely all partitions succeed.


shanson7:
Note: we've been running this in prod for a long time now and have seen a large reduction in startup failures.

robert-milan (approved):

PR #2039, opened 2022-07-29
Reduce AggMetrics size

shanson7:
- Combine 2 interfaces into 1
- Remove unneeded `numChunks` param

With this byte alignment, we should go from 184 bytes (9 padding bytes) to 160 bytes (5 padding bytes). Considering there are millions of these (billions across a cluster) it can add up.

There is also an optimization to only call `Encode` on cacheable elements which reduces the number of allocations / heap spikes during periodic flushes.


shanson7:
Note: we've been running this in production for almost a year (had conflicts with other PRs so couldn't upstream). The results have been good, but as you can imagine it's just a drop in the bucket of the total memory usage.

robert-milan (approved):

PR #2046, opened 2022-12-26
Feature linear regression2 rpm

robert-milan:
Needed to update testing frameworks for e2e tests

### DIFF
--- a/stacktest/docker/docker.go
+++ b/stacktest/docker/docker.go
@@ -189,6 +189,6 @@ func updateEnv(setEnv map[string]string, env []string) []string {
 	return env
 }
 
-func IsDockerCompose() bool {
-	return (docker == "docker-compose")
+func IsNativeDocker() bool {
+	return (docker == "docker")
 }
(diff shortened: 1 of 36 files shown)


stacktest/docker/docker.go
```diff
@@ -156,3 +188,7 @@ func updateEnv(setEnv map[string]string, env []string) []string {
 
 	return env
 }
+
+func IsDockerCompose() bool {
```
shanson7, line 192: This is a bit confusing since it doesn't really convey a difference between `docker-compose` and `docker compose`.  Maybe invert this and make it `IsNativeDocker` or something? I think this would also clean up usage as it seems like the result of this is always negated.

shanson7 (approved):
One minor comment, otherwise LGTM

stacktest/docker/docker.go
robert-milan: Yeah good point, I was in a bit of a hurry. I'll clean it up

PR #2049, opened 2023-06-06
Infinity values no longer produce invalid JSONs

npazosmendez:
Previously they were being serialized as `+Inf`/`-Inf`, which resulted in an unparsable JSON

### DIFF
--- a/api/models/series.go
+++ b/api/models/series.go
@@ -291,7 +291,7 @@ func (series SeriesByTarget) MarshalJSONFast(b []byte) ([]byte, error) {
 		b = append(b, `,"datapoints":[`...)
 		for _, p := range s.Datapoints {
 			b = append(b, '[')
-			if math.IsNaN(p.Val) {
+			if math.IsNaN(p.Val) || math.IsInf(p.Val, 1) || math.IsInf(p.Val, -1) {
 				b = append(b, `null,`...)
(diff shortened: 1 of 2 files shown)


api/models/series.go
```diff
@@ -291,7 +291,7 @@ func (series SeriesByTarget) MarshalJSONFast(b []byte) ([]byte, error) {
 		b = append(b, `,"datapoints":[`...)
 		for _, p := range s.Datapoints {
 			b = append(b, '[')
-			if math.IsNaN(p.Val) {
+			if math.IsNaN(p.Val) || math.IsInf(p.Val, 1) || math.IsInf(p.Val, -1) {
```
leizor, line 294: I think this might work as well:
```suggestion
			if math.IsNaN(p.Val) || math.IsInf(p.Val, 0) {
```

api/models/series.go
```diff
@@ -330,7 +330,7 @@ func (series SeriesByTarget) MarshalJSONFastWithMeta(b []byte) ([]byte, error) {
...
```
leizor, line 333: (ditto)
```suggestion
			if math.IsNaN(p.Val) || math.IsInf(p.Val, 0) {
```

leizor (approved):

